PR #5, opened 2025-10-07
Feature/dn 3294

annesnour03:
https://uvanose.atlassian.net/browse/DN-3294


UvA.Workflow.Api/Features/Actions/ActionsController.cs
```diff
@@ -0,0 +1,48 @@
+using UvA.Workflow.Api.Exceptions;
+using UvA.Workflow.Api.Features.Actions.Dtos;
+
+namespace UvA.Workflow.Api.Features.Actions;
+
+[ApiController]
+[Route("api/actions")]
```
goomens, line 7: I think it can also automatically do this based on the name of the controller. Do we want that? Also do we want "api" everywhere?

Uva.Workflow/WorkflowInstances/WorkflowInstanceService.cs
```diff
@@ -0,0 +1,235 @@
+using System.Linq.Expressions;
+
+namespace Uva.Workflow.WorkflowInstances;
```
goomens, line 3: UvA or Uva?

Uva.Workflow/WorkflowInstances/WorkflowInstanceService.cs
```diff
@@ -0,0 +1,235 @@
...
+        if (string.IsNullOrWhiteSpace(id))
+            throw new ArgumentException("Id is required", nameof(id));
+
+        return await repository.GetByIdAsync(id);
+    }
+
+    /// <summary>
+    /// Gets a workflow instance by ID or throws if not found
+    /// </summary>
+    public async Task<WorkflowInstance> GetByIdRequiredAsync(string id)
```
goomens, line 47: These methods feel a bit silly. Can't you just use the repository directly if you want this?

Uva.Workflow/WorkflowInstances/WorkflowInstanceService.cs
```diff
@@ -0,0 +1,235 @@
...
+
+
+        await repository.CreateAsync(instance);
+        return instance;
+    }
+
+    /// <summary>
+    /// Gets a workflow instance by ID
+    /// </summary>
+    public async Task<WorkflowInstance?> GetByIdAsync(string id)
```
goomens, line 36: Do we want this `Async` suffix everywhere? We don't do it in DN2 currently but a lot of libraries do.  

Also should we pass cancellation tokens everywhere?

Uva.Workflow/Entities/Domain/Inheritance.cs
```diff
@@ -0,0 +1,58 @@
+namespace Uva.Workflow.Entities.Domain;
```
goomens, line 1: Is `Entities` the best name for this?

UvA.Workflow.Api/Features/WorkflowInstances/Dtos/WorkflowInstanceDto.cs
```diff
@@ -0,0 +1,45 @@
+namespace UvA.Workflow.Api.Features.WorkflowInstances.Dtos;
+
+public record WorkflowInstanceDto(
+    string Id,
+    string EntityType,
+    string? Variant,
```
goomens, line 6: I think this Variant thing might not be used anymore and is something left over from an old version?

UvA.Workflow.Api/Exceptions/ErrorCodes/EntityTypesErrorCodes.cs
```diff
@@ -0,0 +1,12 @@
+using System.Net;
+
+namespace UvA.Workflow.Api.Exceptions;
+
+public partial class ErrorCode
+{
+    public static readonly ErrorCode EntityTypesNotFound = new(
```
goomens, line 7: Should not be plural I think?

Uva.Workflow/Services/AnswerConversionService.cs
```diff
@@ -0,0 +1,66 @@
+using Microsoft.AspNetCore.Http;
+
+namespace Uva.Workflow.Services;
+
+public record AnswerInput(
```
goomens, line 5: I think now that we do REST we could change this to just have a single `Value` property of type `JsonElement`? (or maybe we need to handle the files separately)

UvA.Workflow.Api/Features/Submissions/Dtos/SubmissionDto.cs
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    // public Question GetQuestion(ModelService modelService)
+    //     => Question.FromModel(modelService.EntityTypes[EntityType].Properties[QuestionName]);
+    //
+    // public Task<Message[]> GetMessages() // MessagesByObjectIdDataLoader loader
+    //     => Task.FromResult<Message[]>([]); // TODO: loader.LoadAsync(new MessageId(TargetType.Answer, Id));
+}
+
+public record Value(
```
goomens, line 85: This type could also perhaps be replaced by a `JsonElement`? (right now stuff is serialized into `Text` which has sort of necessary in GraphQL but should not be now)

PR #98, opened 2026-03-17
DN-3639: feat: Configure hostnames for Milestones

danielleklaasen:


Deployment/workflow-api/values.yaml
```diff
@@ -2,7 +2,7 @@ hostname: workflow-api.local
 ingressClass: nginx
 repo: datanosecr.azurecr.io
 tag: latest
-certificateName: datanose-nl
+certificateName: milestones-fnwi-uva-nl
```
goomens: @danielleklaasen Do we have a single certificate for tst and acc or separate ones? Does this to be in the values?

danielleklaasen: Yes there are separate ones for tst and acc indeed, this should be the template (is the default value), and is overridden by the gitops value for the helm chart. I adjusted it to always have the tst certificate as default

goomens (approved):

PR #107, opened 2026-04-01
DN-3630: Multiple Assessments

dj-petersen:
Easiest way to test is by using the new project type Project-PP, which includes one Assessment step with three children: AssessmentSupervisor, AssessmentSecondreader and AssessmentExaminer. A new project of this type can be created using the /WorkflowInstances endpoint on swagger.


UvA.Workflow/Assessments/AssessmentService.cs
```diff
@@ -4,30 +4,48 @@ namespace UvA.Workflow.Assessments;
...
+        return actualForm.Pages
+            .Where(p => p.Sources == null || p.Sources.Contains(form.PropertyName))
+            .ToArray();
+    }
+
     public static Dictionary<string, Result[]> CalculateFormResults(SubmissionContext submissionContext,
         string? pageName)
     {
-        int totalWeight = submissionContext.Form.Pages
+        var pages = GetPagesForResults(submissionContext.Form.ActualForm);
```
codure: This looks like a double "Form.ActualForm" step. The function is also doing a
var actualForm = form.ActualForm;

so it's taking 'submissionContext.Form.ActualForm.ActualForm'?

dj-petersen: Well spotted! I've fixed it

dj-petersen: Apparently it worked because it was broken... I removed the filtering here and added it further on so that the percentages keep showing for the base assessment form, but the answers and questions only load for the child form.

danielleklaasen (approved):
Looks like mentioned issues are fixed :) only now there are some merge conflicts unfortunately. So I'll put the ticket in progress, but after that I think it can go to testing immediately, right?

goomens (dismissed):

goomens (approved):

PR #108, opened 2026-04-02
DN-3653 feat(DataNose): Add DataNose service configuration and FinalG…

annesnour03:
…radeSubmission operation

I have added a `template` property to the output of a operation. We needed to create the string which is used to redirect to DN1. This info is kind of only scoped to the service call, so I think it made sense to handle it there. 

I also lended some things from the [TII PR](https://github.com/UvA-FNWI/workflow-api/pull/100), so sorry if there is a bit of overlap between the two PR's

This PR is dependend on its counterpart in https://github.com/UvA-FNWI/UvA.DataNose.WebAPI/pull/1278, I have pointed the DN2 api url to that PR environment so that we can test.
> [!CAUTION]
> Revert [51398f2](https://github.com/UvA-FNWI/workflow-api/pull/108/commits/51398f2c34afc91b1ed7de0ffcc868cf899d31c3) before merging


Examples/Projects/Common/Services/DataNose.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+operations:
+
+- name: FinalGradeSubmission
+  url: Grading/External/CreateProjectGradeList
+  method: POST
+  inputs:
+    - name: CourseExternalId
+      type: String!
+    - name: Student
```
goomens, line 14: Ok I now realise there is an oddity here... we don't really have the student ID anywhere, only the username. That is almost always the same so maybe that is ok, but let's think about this for a moment

goomens (dismissed):

Examples/Projects/Common/Services/DataNose.yaml
annesnour03: Yeah we only have the `username`, does this ever differ? I know in tst we have `invalidXXXXX`, but I assumed that in acc/prd it would be the same

goomens: In 99% of the cases yes...

goomens (dismissed):

goomens (approved):

PR #109, opened 2026-04-02
feature/DN-3667

codure:
Added multiple role checks for multiple controllers.
The most important ones:
- AssessmentsController
- StepsController
- AnswersController
- SubmissionsController
- UsersController
- VersionsController


UvA.Workflow/Jobs/JobService.cs
```diff
@@ -111,7 +111,9 @@ private async Task<EffectResult> RunJob(Job job, WorkflowInstance instance, Effe
 #if DEBUG
                 throw;
 #endif
+#pragma warning disable CS0162 // Unreachable code detected
```
annesnour03: This is reachable in non-debug environments right? I think we can remove this `#pragma`

UvA.Workflow.Api/Infrastructure/ApiControllerBase.cs
```diff
@@ -1,6 +1,7 @@
 using System.Net;
 using Microsoft.AspNetCore.Authorization;
 using UvA.Workflow.Api.Authentication;
+using UvA.Workflow.Infrastructure;
```
annesnour03: minor: can be removed

UvA.Workflow.Api/Submissions/SubmissionsController.cs
```diff
@@ -19,6 +20,9 @@ public async Task<ActionResult<SubmissionDto>> GetSubmission(string instanceId,
     {
         var (instance, submission, form, _) =
             await submissionService.GetSubmissionContext(instanceId, submissionId, version, ct);
+
+        await rightsService.EnsureAuthorizedForAction(instance, RoleAction.View);
```
annesnour03: I think this is too broad, as `rightsService.EnsureAuthorizedForAction(instance, RoleAction.View);` returns true if we can view **any** submission on the form. Should we scope the rights check to the submission that we're requesting? So I think scope to `form.Name`?

Maybe we can also account for this in the unit test?

UvA.Workflow.Api/Submissions/SubmissionsController.cs
```diff
@@ -28,12 +32,16 @@ public async Task<ActionResult<SubmissionDto>> GetSubmission(string instanceId,
...
-        if (currentUser == null)
+        var user = await userService.GetCurrentUser(ct);
+        if (user == null)
             return Unauthorized();
+
         var context = await submissionService.GetSubmissionContext(instanceId, submissionId, null, ct);
         var (instance, sub, form, _) = context;
-        var result = await submissionService.SubmitSubmission(context, currentUser, ct);
+
+        await rightsService.EnsureAuthorizedForAction(instance, RoleAction.Submit);
```
annesnour03: I think this has the same scoping bug right?

UvA.Workflow.Tests/Controllers/UsersControllerTests.cs
```diff
@@ -0,0 +1,61 @@
...
+    [InlineData("Admin")]
+    [InlineData("Api")]
+    public async Task Users_GetLoggedInUser_Allow(string role)
+    {
+        // Arrange
+        var controller = BuildControllerWithRoles([role]);
+        // Act
+        var result = await controller.GetLoggedInUser(_ct);
+        //Assert
+        Assert.IsType<ActionResult<UserDto>>(result);
```
annesnour03: This is always going to be true right? The returntype of `GetLoggedInUser` is always going to be in this form, regardless whether the user is actually logged in or not. Should we check the content of `result`?

This same check is done here in this file, but also in `SubmissionsController`, `StepsController` and in `ScreensController`, should we fix that?

annesnour03 (changes requested):
Mainly looks good!

I had a few comments and a suggestion about simplifying the tests, maybe because we already have the `ControllerTestBase`, we can add some mocks that are used in the controllers:
```cs
protected void MockCurrentUser(params string[] roles)
{
    _userServiceMock.Setup(s => s.GetRolesOfCurrentUser(It.IsAny<CancellationToken>()))
        .ReturnsAsync(roles);
    _userServiceMock.Setup(s => s.GetCurrentUser(It.IsAny<CancellationToken>()))
        .ReturnsAsync(ControllerTestsHelpers.AdminUser);
}

protected void MockInstance(WorkflowInstance instance)
{
    _workflowInstanceRepoMock.Setup(r => r.GetById(instance.Id, It.IsAny<CancellationToken>()))
        .ReturnsAsync(instance);
}

protected void MockEmptyRelatedInstanceLookups()
{
    _workflowInstanceRepoMock.Setup(r => r.GetAllById(It.IsAny<string[]>(),
            It.IsAny<Dictionary<string, string>>(),
            It.IsAny<CancellationToken>()))
        .ReturnsAsync([]);
}

protected void MockEmptyEventLog(WorkflowInstance instance)
{
    _eventRepoMock.Setup(r => r.GetEventLogEntriesForInstance(instance.Id,
            It.IsAny<List<string>>(),
            It.IsAny<CancellationToken>()))
        .ReturnsAsync([]);
}
```

UvA.Workflow/Jobs/JobService.cs
codure: I've replaced it with a simpler "#else"

UvA.Workflow.Tests/Controllers/UsersControllerTests.cs
codure: Fair enough, it's not really verifying anything other than returning 'something'.
I'll add some simple additional checks here and there to maybe get the ball rolling.

UvA.Workflow.Api/Submissions/SubmissionsController.cs
codure: I've added the form name to the check(s) 👍

UvA.Workflow.Api/Submissions/SubmissionsController.cs
codure: I've added the form name to the check, and updated the unit test accordingly.

UvA.Workflow.Tests/Controllers/UsersControllerTests.cs
annesnour03: Yeah they always succeeded. I noticed that they're still present in `SubmissionsController`, `StepsController` and in `ScreensController`, should we fix that?

codure: I've updated the tests in all of them now. 👍

goomens (approved):

PR #111, opened 2026-04-07
Feature/dn 3670

annesnour03:
https://uvanose.atlassian.net/browse/DN-3670
### EduID
The main change is that EduID behavior is split behind dedicated interfaces and services (`IEduIdUserService`, invitation client, role/search sources) and then plugged into the existing pipeline through DI. `SurfConextAuthenticationHandler` now only decides whether a login is EduID or internal and delegates the rest, while `UserService` no longer knows about one specific external provider implementation. Instead, it consumes generic `IUserRoleSource` and `IUserSearchSource` extensions, which keeps user roles/search loosely coupled.

To track the authentication state of an external user, the `User` model has two extra flags, one to indicate the auth provider, and one to track whether the login is already active or not.
### Surf
To properly connect the new eduid role, this PR also introduces the new surf clientid. To make this change final we need to also change it in the gitops repo for all tst environments. We can't do this yet, as there will be a mismatch between clientid/secret.

### Tests
Apparantly we were not running tests on builds? I'm a bit confused about this, but I had non building tests when I pulled from main. I fixed those tests and also added a test step in the dockerfile.

This PR also uses https://github.com/UvA-FNWI/SecureEduIdInviteProxy/pull/3.


Deployment/workflow-api-pr/templates/configmap-api.yaml
```diff
@@ -19,6 +19,14 @@ data:
 
   SurfConext__BaseUrl: "{{ .Values.surfConext.baseUrl }}"
   SurfConext__ClientId: "{{ .Values.surfConext.clientId }}"
+  EduId__Authority: "{{ .Values.eduId.authority }}"
+  EduId__InvitationApiUrl: "{{ .Values.eduId.invitationApiUrl }}"
+  EduId__RoleIdentifier: "{{ .Values.eduId.roleIdentifier }}"
+  EduId__InvitationExpiryDays: "{{ .Values.eduId.invitationExpiryDays }}"
+  EduId__RoleExpiryDays: "{{ .Values.eduId.roleExpiryDays }}"
+  EduId__InternalEmailDomains__0: "{{ index .Values.eduId.internalEmailDomains 0 }}"
```
goomens: Maybe better to do a single env variable with comma-separated list here? Or is that uglier? (or can we do this with some kind of Helm loop)

Deployment/workflow-api-pr/values.yaml
```diff
@@ -67,8 +79,10 @@ azure:
       key: DatanoseApi__ApiKey
     - name: workflow-encryption-secret
       key: Encryption__Secret
-    - name: workflow-surfconext-client-secret
+    - name: workflow-surfconext-client-secret-v2
```
goomens, line 82: What's going on here?

UvA.Workflow.Api/Authentication/EduIdUserService.cs
```diff
@@ -0,0 +1,182 @@
...
+    ILogger<EduIdUserService> logger) : IEduIdUserService
+{
+    private readonly EduIdOptions _options = options.Value;
+
+    public bool IsInternalEmailAddress(string email)
+    {
+        var trimmedEmail = email.Trim();
+        var atIndex = trimmedEmail.LastIndexOf('@');
+        if (atIndex < 0 || atIndex == trimmedEmail.Length - 1)
+            return false;
```
goomens: This could also throw an exception I guess 😛

UvA.Workflow.Api/Infrastructure/ServiceCollectionExtentions.cs
```diff
@@ -42,6 +44,19 @@ public static IServiceCollection AddWorkflow(this IServiceCollection services, I
         services.AddScoped<IUserRepository, UserRepository>();
         services.AddScoped<IInstanceEventRepository, InstanceEventRepository>();
         services.AddScoped<IJobRepository, JobRepository>();
+        services.AddScoped<EduIdUserDirectory>();
```
goomens: Maybe here we could separate things out between core components of the workflow system and additions like this that add specific types of authentication?

UvA.Workflow.Api/Invites/InvitesController.cs
```diff
@@ -0,0 +1,38 @@
+using UvA.Workflow.Api.Authentication;
+using UvA.Workflow.Api.Infrastructure;
+
+namespace UvA.Workflow.Api.Invites;
+
+/// <summary>
+/// Creates a pending EduID user and EduID invitation via <see cref="IEduIdUserService"/>.
+/// </summary>
+public class InvitesController(IEduIdUserService eduIdUserService) : ApiControllerBase
```
goomens, line 10: Is this a temporary thing for testing or are we going to need this in the future?

UvA.Workflow/Users/User.cs
```diff
@@ -16,4 +16,10 @@ public class User
     [BsonElement("DisplayName")] public string DisplayName { get; set; } = null!;
 
     [BsonElement("Email")] public string Email { get; set; } = null!;
+
+    [BsonElement("AuthProvider")]
+    [JsonIgnore]
+    public UserAuthProvider AuthProvider { get; set; } = UserAuthProvider.Internal;
```
goomens, line 22: Question: are we also storing this `User` object in the `User` properties of instances? If so maybe we should separate that out so we don't store stuff like this there?

UvA.Workflow/Users/UserAuthProvider.cs
```diff
@@ -0,0 +1,7 @@
+namespace UvA.Workflow.Users;
+
+public enum UserAuthProvider
+{
+    Internal = 0,
```
goomens, line 5: `Internal` is basically everything else? Both UvA and Amsterdam UMC?

UvA.Workflow/Users/UserService.cs
```diff
@@ -106,12 +116,18 @@ public class UserService(
...
+        if (_cache.TryGetValue(cacheKey, out string[]? roles)) return roles!;
         if (user.UserName == ApiUserName)
             roles = ["Api"];
         else
-            roles = (await dataNoseApiClient.GetRolesByUser(user.UserName, ct)).ToArray();
-        cache.Set(cacheKey, roles, RolesCacheExpiration);
+        {
+            var roleSource = _userRoleSources.FirstOrDefault(source => source.CanResolve(user));
+            roles = roleSource == null
+                ? (await dataNoseApiClient.GetRolesByUser(user.UserName, ct)).ToArray()
```
goomens: Why isn't this in a role source as well?

UvA.Workflow/Users/UserService.cs
```diff
@@ -127,5 +143,30 @@ public async Task<IEnumerable<string>> GetRolesOfCurrentUser(CancellationToken c
     }
 
     public async Task<IEnumerable<UserSearchResult>> FindUsers(string query, CancellationToken ct)
-        => await dataNoseApiClient.SearchPeople(query, ct);
+    {
+        var internalUsers = (await dataNoseApiClient.SearchPeople(query, ct)).ToList();
```
goomens: And this?

.dockerignore
```diff
@@ -19,7 +19,8 @@
 **/node_modules
 **/npm-debug.log
 **/obj
+UvA.Workflow.Api/data
```
goomens, line 22: What is this?

UvA.Workflow.sln
```diff
@@ -22,6 +22,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "SolutionItems", "SolutionIt
 	EndProjectSection
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "UvA.Workflow.Tests", "UvA.Workflow.Tests\UvA.Workflow.Tests.csproj", "{F569D8D8-7C9F-422F-8C6A-933C00E5FF05}"
```
goomens: Random aside: there is a new `slnx` file format for this thing that is much more readable than this odd guid thing

goomens (commented):
One general comment: this is all a bit specific stuff that I guess that at some point we want to be able to decouple from the workflow system itself. That's quite doable with this setup and the source thingies, I think? (and we can also maybe decouple DN like that?)

UvA.Workflow.sln
annesnour03: Oh that's a lot better, I've changed it now

.dockerignore
annesnour03: The `data` folder is MongoDB's volume that it mounts to, when running build in docker locally we don't really want to take this with us.

Deployment/workflow-api-pr/values.yaml
annesnour03: Well, we have a new clientid/secret for milestones. I wanted to use that pair in this PR, but I can't update the old `workflow-surfconext-client-secret` to just be that new secret, because auth in all other PR's will be broken then. So I've created a 2nd tmp secret, when this PR is merged we can remove it and update the old one I guess?

UvA.Workflow.Api/Invites/InvitesController.cs
annesnour03: This is a testing thing, I don't think we are going to need this in the future, I think we will do it with https://uvanose.atlassian.net/browse/DN-3732.

Deployment/workflow-api-pr/templates/configmap-api.yaml
annesnour03: I think the helm loop will be nicer here, I've implemented that

UvA.Workflow/Users/UserAuthProvider.cs
annesnour03: It currently is yes, should we differentiate between UvA and Amsterdam UMC?

goomens: I think it might not be needed

UvA.Workflow/Users/User.cs
annesnour03: I have now created an `InstanceUser`, it did came with a bit of a big change, so see that in my commit

UvA.Workflow/Users/UserService.cs
annesnour03: Kind of treated DN as it's own thing, but I guess there is no reason why its not a source like the rest. I have now refactored this

UvA.Workflow/Users/UserService.cs
annesnour03: fixed in [c6faa22](https://github.com/UvA-FNWI/workflow-api/pull/111/commits/c6faa221c8f016bacbf797d52c1e3950b641b3c6)

UvA.Workflow.Api/Infrastructure/ServiceCollectionExtentions.cs
annesnour03: Yeah this is a good point, for now I've just grouped it into its own function, but this is not really enough. I've created a PR #120 which is a big and more strict refactor of what is core logic and what belongs in api/misc. Maybe this is something we can look at together? This kind of goes hand in hand with your overall comment

UvA.Workflow/Users/InstanceUser.cs
```diff
@@ -0,0 +1,24 @@
+using MongoDB.Bson.Serialization.Attributes;
+
+namespace UvA.Workflow.Users;
+
+public class InstanceUser
```
goomens, line 5: Maybe not for this PR, but I guess we need to add organisation here so we can store that? And maybe the ID can be optional so we can also store external people for which there is no user yet?

goomens (approved):

goomens (dismissed):

goomens (approved):

PR #115, opened 2026-04-14
DN-3737 feat(toast): add showToast effect

annesnour03:
https://uvanose.atlassian.net/browse/DN-3737

### DIFF
--- a/Examples/Projects/Project-PA/Steps/Start.yaml
+++ b/Examples/Projects/Project-PA/Steps/Start.yaml
@@ -10,6 +10,17 @@ actions:
   - roles: [Student]
     type: Submit
     form: Start
+  - roles: [Student]
+    type: Execute
+    label:
+      en: Test toast
+      nl: Test toast
+    onAction:
+      - showToast:
+          type: Success
+          message:
+            en: Project proposal saved successfully
+            nl: Projectvoorstel succesvol opgeslagen
 
 properties:
   - name: ProjectLocation
--- a/Examples/Projects/Project-PA/Steps/Start.yaml
+++ b/Examples/Projects/Project-PA/Steps/Start.yaml
@@ -10,17 +10,6 @@ actions:
   - roles: [Student]
     type: Submit
     form: Start
-  - roles: [Student]
-    type: Execute
-    label:
-      en: Test toast
-      nl: Test toast
-    onAction:
-      - showToast:
-          type: Success
-          message:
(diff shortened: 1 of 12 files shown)


UvA.Workflow/WorkflowInstances/EffectService.cs
```diff
@@ -13,10 +13,16 @@
 
 namespace UvA.Workflow.WorkflowInstances;
 
-public record EffectResult(string? RedirectUrl = null, bool? ShowConfetti = null)
+public record ShowToastResult(ToastType Type, BilingualString Message);
+
+public record EffectResult(string? RedirectUrl = null, bool? ShowConfetti = null, ShowToastResult? ShowToast = null)
```
dj-petersen: It's a little confusing that ShowConfetti is a boolean and ShowToastResult is an object. Maybe renaming it simply to `ToastResult? Toast` would be clearer.

Examples/Projects/Project/Steps/Start.yaml
```diff
@@ -11,6 +11,18 @@ actions:
...
+    type: Execute
+    name: TestToast
+    label:
+      en: Test toast
+      nl: Test toast
+    onAction:
+      - showToast:
+          type: Success
+          message:
+            en: "Project submitted successfully"
```
dj-petersen: Maybe we can change the test data to the toast message in [this figma design](https://www.figma.com/design/nRpDt5pm6DElQCI20x5inG/Scriptieworkflow?node-id=4325-9299&t=v6z91iYNF9StZLRE-1)? 
Then there is not a separate button needed for the toast

Examples/Projects/Project-PA/Steps/Start.yaml
```diff
@@ -10,6 +10,17 @@ actions:
   - roles: [Student]
     type: Submit
     form: Start
+  - roles: [Student]
+    type: Execute
+    label:
```
dj-petersen: Name is missing here, which gives an 400 error in the api

danielle-devv (changes requested):
Looks good! I have some minor proposed changes.

Examples/Projects/Project/Steps/Start.yaml
annesnour03: Sure!

danielle-devv (approved):
Nice!

PR #116, opened 2026-04-14
DN-3727 feat(parser): parse layout booleans correctly

annesnour03:
https://uvanose.atlassian.net/browse/DN-3727

FE got `"multiline" : "true"` instead of an actual boolean


goomens (approved):
Wacky that the yaml library doesn't do this

PR #118, opened 2026-04-16
POC DN-3700 feat(project): create first version for RMSS

annesnour03:
https://uvanose.atlassian.net/browse/DN-3700

This is a POC for the RMSS proposal. It includes a few simple questions and a double approval step (supervisor and proposalreviewer). In the future we want to add a rubric here.


Examples/Projects/Project-RMSS/Entity.yaml
```diff
@@ -0,0 +1,34 @@
+name: Project-RMSS
+title:
+  en: Project (RMSS)
+  nl: Project (RMSS)
+titlePlural:
+  en: Projectss (PA)
```
goomens: Paaaa

Examples/Projects/Project-RMSS/Properties.yaml
```diff
@@ -0,0 +1,31 @@
...
+    type: String!
+    text:
+      en: Working title
+      nl: Werktitel
+
+  - name: EC
+    type: Int!
+    text:
+      en: EC
+      nl: EC
```
goomens: Maybe for this basic stuff we can use inheritance?

Examples/Projects/Project-RMSS/Properties.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+  - name: ProposalReviewer
```
goomens, line 6: Who is this in practice?

Examples/Projects/Project-RMSS/Steps/Start.yaml
```diff
@@ -0,0 +1,108 @@
...
+    validation:
+      message:
+        en: End date must be after start date
+        nl: Einddatum moet na de startdatum zijn
+      value:
+        property: PlannedEndDate
+        greaterThan: StartDate
+
+  - name: Interruptions
+    type: Date
```
goomens, line 100: This one is currently a bit odd

Examples/Projects/Project-RMSS/Steps/ReviewSupervisor.yaml
```diff
@@ -0,0 +1,70 @@
...
+    form: ThesisProposalReview
+    label:
+      en: Submit review
+      nl: Beoordeling indienen
+
+properties:
+  - name: SupervisorResearchProposalSufficient
+    type: SupervisorResearchProposalSufficientType!
+    text:
+      en: Is the proposal sufficient?
```
goomens: For these kind of specific forms I think we need to think a little bit about language. Is RMSS a bilingual programme?

Examples/Projects/Project-RMSS/Steps/ReviewProposalReviewer.yaml
```diff
@@ -0,0 +1,70 @@
+name: ReviewProposalReviewer
```
goomens, line 1: For these two forms, should we use the same approach as the multiple assessment and have an embedded entity instead of duplicating the properties?

Examples/Projects/Project-RMSS/Screens/Projects-PA.yaml
```diff
@@ -0,0 +1,19 @@
+name: Projects-PA
```
goomens: Paaa?

Examples/Projects/Project-RMSS/Forms/ThesisProposalReview.yaml
```diff
@@ -0,0 +1,45 @@
+name: ThesisProposalReview
+title:
+  en: Research proposal and ethics review
+  nl: Onderzoeksvoorstel en ethiek beoordeling
+
+onSubmit:
+  - condition:
```
goomens, line 12: I think ideally we'd avoid effects that set events because there's some risk of getting into odd states. Would there be another way to handle this, e.g. with steps just ending based on the right condition?

Examples/Projects/Project-RMSS/Properties.yaml
annesnour03: In practice it's the course coordinator. Sometimes it's one coordinator that does all the students, and sometimes there are 2 coordinators that share the load

Examples/Projects/Project-RMSS/Screens/Projects-PA.yaml
annesnour03: oops

Examples/Projects/Project-RMSS/Properties.yaml
annesnour03: Yes we can!

Examples/Projects/Project-RMSS/Steps/ReviewProposalReviewer.yaml
annesnour03: I've done this now :)

Examples/Projects/Project-RMSS/Forms/ThesisProposalReview.yaml
annesnour03: So when a timeline is linear I think we can do this, because we don't really need to set/fire events. But now that we also have the reject option, we need to somehow fire an event so that we can suppress it, but I don't think the `ends` supports this?

Examples/Projects/Project-RMSS/Properties.yaml
goomens: The course coordinator role is already there from the context so I guess it doesn't need to be linked at the project level?

Examples/Projects/Project-RMSS/Forms/ThesisProposalReview.yaml
goomens: Oof... maybe we can think about this together tomorrow/Thursday

goomens (dismissed):

Examples/Projects/Project-RMSS/Forms/ThesisProposalReview.yaml
goomens: Ah I forget about this one. I think the current solution has the disadvantage that you now have one action (submission of a form), generating two events: the actual submission event and the one that follows from this. Then if you start undoing stuff it might get confusing.

Would a possible solution be to allow dynamic events on form submission? So submitting the form can generate either the one event or the other? And both events make the form count as 'submitted'.

(The alternative would be do something with condition suppression but that might be complete madness?)

UvA.Workflow/WorkflowModel/Form.cs
```diff
@@ -136,6 +136,17 @@ public class Form : INamed
     /// </summary>
     public Effect[] OnSubmit { get; set; } = [];
 
+    /// <summary>
+    /// Whether submitting the form should emit the form-name event.
+    /// </summary>
+    public bool EmitFormSubmitEvent { get; set; } = true;
```
goomens, line 142: Isn't this a bit superfluous? If `SubmittedWhenEvents` is filled, this should always be false, I think?

goomens (dismissed):

UvA.Workflow/WorkflowModel/Form.cs
annesnour03: Yes this is true, when `SubmittedWhenEvents` is filled in we basically always want to have `EmitFormSubmitEvent` set to true. The reason I added it though, was because I felt having `SubmittedWhenEvents` decide when its submitted __and__ not setting the submit event was maybe a bit too much for one option. Felt a bit like a side effect, what do you think?

goomens: Yes, maybe you're right. Let's try it like this for a bit

goomens (approved):

PR #119, opened 2026-04-16
DN-3610: First Emails

dj-petersen:
For this ticket I have added four emails:

- SubjectSubmitted ([design](https://www.figma.com/design/nRpDt5pm6DElQCI20x5inG/Scriptieworkflow?node-id=8260-135175&t=QcjbZGJKXNr4yIz7-1)) 
- FinalVersionSubmitted ([design](https://www.figma.com/design/nRpDt5pm6DElQCI20x5inG/Scriptieworkflow?node-id=8260-158334&t=QcjbZGJKXNr4yIz7-1 ))
- AssessmentReminderExaminer ([design](https://www.figma.com/design/nRpDt5pm6DElQCI20x5inG/Scriptieworkflow?node-id=8260-160318&t=QcjbZGJKXNr4yIz7-1) )
- AssessmentRequestAssessor ([design](https://www.figma.com/design/nRpDt5pm6DElQCI20x5inG/Scriptieworkflow?node-id=8260-161164&t=QcjbZGJKXNr4yIz7-1 ))

**To the following workflows/steps:**

_Project/Start:_ 
- SubjectSubmitted to Student
- AssessmentRequestAssessor to SecondReviewer (to test the missing values fallback.)

_Project/Upload:_
- FinalVersionSubmitted to Student

_Project-PA/Start:_
- SubjectSubmitted to Student

_Project-PA/UploadFinalThesis:_
- FinalVersionSubmitted to Student
- AssessmentReminderExaminer to Examiner

_Project-PA/AssessmentExaminer:_
- AssessmentRequestAssessor to SecondReviewer 

I've also changed the Student type from [User]! to User! which means that a **migration** is needed on the test database and local databases after merging. It also means that **this PR can only be tested with new workflows.** 

The migration script that can be run in the DataGrip MongoDB console:
```
use workflow;

    const withStudentArray = db.instances.find({
        "Properties.Student": {$type: "array"}
        }).toArray();

    print(withStudentArray)

    for (const row of withStudentArray){
        const arr = row.Properties.Student;
        // Safety check: ensure it's an array with at least one element
        if (arr && arr.length && typeof arr.length === "number") {

            // Build dynamic update object
            const update = {
                "Properties.Student": arr[0]
                };

            db.instances.updateOne(
                { _id: row._id },
                { $set: update }
                );

            print("Updated instance with _id: ");
            print(row._id.valueOf())
            }
        }
```


Examples/Projects/Project-PA/Emails/AssessmentReminderExaminer.yaml
```diff
@@ -0,0 +1,36 @@
+name: AssessmentReminderExaminer
+layout: default
+toAddress: "replaced-by-override@mail.com"
```
goomens: Do we already have stuff in place so we can refer to do examiner here?

UvA.Workflow/Notifications/MailBuilder.cs
```diff
@@ -1,42 +1,84 @@
...
-        return new MailMessage(subject, fullHtml) { To = [recipient] };
+        return Task.FromResult(new MailMessage(subject, fullHtml) { To = [recipient] });
+    }
+
+    private static SendMessage ResolveMessageContents(WorkflowDefinition workflowDefinition, SendMessage sendMail)
+    {
+        if (string.IsNullOrWhiteSpace(sendMail.TemplateKey))
+            return sendMail;
+
+        var template = workflowDefinition.Emails
```
annesnour03, line 50: This now only looks inside the `Emails` folder to search for templates, I think this doesn't work for seeded workflows, like the one in `SubsidyNew`. An example of this is the `DecisionMail`. Should we still support this?

We could also decide that we just don't support that anymore if we don't think that's a good idea, and only have emails in the `Emails` folder, but then we just need to migrate that workflow over.

Examples/Projects/Project/Emails/AssessmentRequestAssessor.yaml
```diff
@@ -0,0 +1,34 @@
...
+  nl: "Geef aan of je het eens bent met de beoordeling voor de scriptie van {{ coalesce(Student.DisplayName, =deze student) }}"
+body:
+  en: |
+    Dear {{ coalesce(Reviewer.DisplayName, =colleague) }},
+    
+    The thesis of **{{ Student.DisplayName }}** has been assessed by the supervisor.<br>
+    You are the second reader for this thesis.
+   
+    Could you please let us know as soon as possible whether you agree with this assessment or not?
+    Please provide this **before {{ coalesce(formatDate(addWeeks(UploadFinalThesisEvent, 2), =dd-MM-yyyy), =the deadline) }}**. That is in {{ coalesce(daysUntil(now, addWeeks(UploadFinalThesisEvent, 2)), =a few) }} days.
```
annesnour03: I don't know how much this matters, but I don't think we have a `UploadFinalThesis` in the `Project` workflow right? I see that it is present in PA

UvA.Workflow/Notifications/MailBuilder.cs
dj-petersen: For me it's most logical to have Emails for a specific workflow in the workflow's folder. But maybe @goomens  has an opinion on this?

Examples/Projects/Project/Emails/AssessmentRequestAssessor.yaml
dj-petersen: Nice catch! I see the event is called Upload in this workflow. I will change it.

Examples/Projects/PA/Project-PA/Emails/AssessmentReminderExaminer.yaml
```diff
@@ -0,0 +1,36 @@
+name: AssessmentReminderExaminer
+layout: default
+toAddress: "replaced-by-override@mail.com"
```
goomens: Shouldn't this have the actual role now?

Examples/Projects/PA/Project-PA/Properties.yaml
```diff
@@ -4,7 +4,7 @@ properties:
     inheritedRoles: [Coordinator]
 
   - name: Student
-    type: "[User]!"
+    type: User!
```
goomens, line 7: Maybe we need to think about this... there are some projects with multiple students. But not many. So what is a good approach here?

UvA.Workflow/Notifications/MailBuilder.cs
goomens: I think maybe we should indeed decide that emails are in yaml and are not data. Although you could probably still refer to some entity in your template if you wanted to?

Examples/Projects/PA/Project-PA/Properties.yaml
goomens: Discussed: for now we just get rid of the array

Examples/Projects/PA/Project-PA/Emails/AssessmentReminderExaminer.yaml
dj-petersen: The recipient is resolved in the form on the onSubmit. the To property trumps the ToAddress. I've removed all the 'override' email addresses.
```
  - sendMail:
      template: AssessmentRequestAssessor
      to: Reviewer
```

annesnour03 (dismissed):

goomens (approved):

PR #120, opened 2026-04-17
DN-3765 refactor(app): restructure more strictly into core/api

annesnour03:
This is a POC based off of the EduID PR, showing a more strict separation between core logic and api logic. I've kind of let AI do its thing with regard to moving things over, I think this could be something.


goomens (commented):
Oof this is big 😊. But it looks good! I think this overdoing it slightly in terms of the number of different projects, maybe? But the overall approach I think is what we want here. Let's discuss on Monday

codure:
The only thing I would recommend is to bundle the Authentications in one project, and bundle the Users in one. the rest is looking great!

goomens (dismissed):

codure (dismissed):
Some minor merge conflicts left, but looking good! Works locally so I'll believe this will work just fine. just needs a good timing for all the other PR's and projects to also update to the new(er) structurel.

goomens (approved):

PR #121, opened 2026-04-21
DN-3677 feat(step): add header status configuration and resolver for …

annesnour03:
…step events
https://uvanose.atlassian.net/browse/DN-3677

### DIFF
--- a/UvA.Workflow/WorkflowModel/Step.cs
+++ b/UvA.Workflow/WorkflowModel/Step.cs
@@ -47,7 +47,7 @@ public class Step : INamed
 
     [YamlIgnore] public Step[] Children { get; set; } = [];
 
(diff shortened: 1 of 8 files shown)


UvA.Workflow/WorkflowModel/Step.cs
```diff
@@ -39,6 +47,8 @@ public class Step : INamed
 
     [YamlIgnore] public Step[] Children { get; set; } = [];
 
+    [YamlMember(Alias = "headerStatus")] public List<StepHeaderStatusConfiguration>? HeaderStatus { get; set; }
```
goomens: Why the alias?

goomens (dismissed):

UvA.Workflow/WorkflowModel/Step.cs
annesnour03: oops I had a different name previously, removed now

codure (approved):

PR #122, opened 2026-04-22
Dn-3748: Inheritance of assessments causes children to mix content

dj-petersen:
The fix prepends the parent’s pages to the child’s pages in the form. It should work independently of the name of the form.


UvA.Workflow/WorkflowModel/ModelParser.cs
```diff
@@ -163,8 +163,11 @@ private void PreProcess(Form form, WorkflowDefinition workflowDefinition)
         workflowDefinition.Events.Add(new() { Name = form.Name });
 
         if (form is { PropertyName: not null, TargetFormName: not null })
-            form.TargetForm = WorkflowDefinitions[workflowDefinition.Properties.Get(form.PropertyName).UnderlyingType]
-                .Forms.Get(form.TargetFormName);
+        {
+            var targetDef = WorkflowDefinitions[workflowDefinition.Properties.Get(form.PropertyName).UnderlyingType];
+            form.TargetForm = targetDef.Forms.GetOrDefault(form.TargetFormName)
+                              ?? targetDef.Forms.FirstOrDefault();
```
goomens: I'm not sure I understand why we have this fallback here?

UvA.Workflow/WorkflowModel/Inheritance.cs
```diff
@@ -8,12 +8,19 @@ private void ApplyInheritance(WorkflowDefinition target, WorkflowDefinition sour
...
-            else
-                target.Forms.Add(sourceForm);
+            foreach (var sourceForm in source.Forms)
+                target.Forms.Add(sourceForm.Clone());
+        }
+        else
+        {
+            foreach (var targetForm in target.Forms)
+            {
+                var parentFormName = target.InheritsFrom ?? targetForm.Name;
```
goomens: Shouldn't this just always be `targetForm.Name`?

UvA.Workflow/Submissions/AnswerService.cs
```diff
@@ -69,7 +69,7 @@ await instanceJournalService.LogPropertyChange(instance.Id,
         var updates = modelService.GetQuestionStatus(instance, form, canViewHidden, questionsToUpdate);
 
         // Build response
-        return Answer.Create(instance, form.TargetForm ?? form, updates);
+        return Answer.Create(instance, form, updates);
```
goomens, line 72: Why is this not (no longer?) needed?

UvA.Workflow/WorkflowModel/ModelParser.cs
dj-petersen: When the child form has another name than the parent form (for example: parent = Assessment, child = Assessment-AI) the first returns to null.

UvA.Workflow/WorkflowModel/Inheritance.cs
dj-petersen: This is also for the case that the child form name is not the same as the parent form name. In that case, the parent form (Assessment) is not added to target.Forms, but the child form is (Assessment-AI). To prepend the questions from the parent form into the child form it checks the inheritsFrom to find the parent form and doesn't depend on them having the same name.

UvA.Workflow/Submissions/AnswerService.cs
dj-petersen: Even when there is inheritance, the form is finally saved under the child's name. Leaving this in gives errors because the form.TargetForm does not exist on the instance. 
(I cherry-picked this from the multiple-assessment's PR)

UvA.Workflow/WorkflowModel/ModelParser.cs
goomens: But isn't that okay? If the form name doesn't match that it shouldn't add a reference, I guess

dj-petersen: I've changed it back. The InheritsFrom property on non-matching names fixes it

UvA.Workflow/WorkflowModel/Inheritance.cs
dj-petersen: I've added a targetForm.InheritsFrom property to match on if the name does not match

goomens (approved):

PR #123, opened 2026-04-23
DN-3734: Allow steps to be conditionally visible

dj-petersen:
The checkbox is added to the Upload step of the Project workflow definition, which shows the publication step conditionally.


UvA.Workflow.Api/WorkflowInstances/Dtos/WorkflowInstanceDtoFactory.cs
```diff
@@ -30,6 +31,7 @@ public async Task<WorkflowInstanceDto> Create(WorkflowInstance instance, Cancell
             RoleAction.ViewAdminTools,
             RightsEvaluationMode.RealUser);
         var viewerRoles = await rightsService.GetViewerRoles(instance, ct);
+        var context = modelService.CreateContext(instance);
```
annesnour03, line 34: Do we need to enrich this context before using it?

UvA.Workflow/Submissions/Answer.cs
```diff
@@ -75,6 +75,15 @@ private static Answer Create(Form form, string questionName,
             );
         }
 
+        if (question.DataType == DataType.Boolean)
+        {
+            // Default to false, not to null
+            var value = ObjectContext.GetValue(answer, question) as bool?;
```
annesnour03, line 86: For hidden questions the value will automatically be `false`, even if we filled them in as `true`. So if we have a boolean that is hidden, but we previously filled that in as `true`, it will be sent to the frontend as `false`. Now that question won't get shown, but it does clutter/mislead the api contract.

Should we just short circuit in the `Create` for all types, so we don't have this issue?

UvA.Workflow.Api/WorkflowInstances/Dtos/WorkflowInstanceDtoFactory.cs
dj-petersen: I don't know if there will be a condition where it is needed in the future, but I added the enrichment to be sure.

UvA.Workflow/Submissions/Answer.cs
dj-petersen: I've added the short circuit for the !isVisible

annesnour03 (approved):

PR #126, opened 2026-04-29
Feature/dn 3732

annesnour03:
https://uvanose.atlassian.net/browse/DN-3732

We still have to find a good way to store the users to make it work with the eduid flow. This is probably something we should discuss.


UvA.Workflow/WorkflowInstances/EffectService.cs
```diff
@@ -122,6 +125,65 @@ private async Task SetProperty(WorkflowInstance instance, ObjectContext context,
...
+                $"External account effect role '{effect.Role}' could not be resolved to workflow users.")
+        };
+
+        var normalizedRecipients = recipients
+            .Select(r => (Recipient: r, Email: r.Email?.Trim() ?? string.Empty))
+            .DistinctBy(r => r.Email, StringComparer.OrdinalIgnoreCase);
+
+        foreach (var (recipient, email) in normalizedRecipients)
+        {
+            if (string.IsNullOrWhiteSpace(email))
```
goomens: I guess this is also handled by the check below 😊

goomens (dismissed):

UvA.Workflow/WorkflowInstances/EffectService.cs
annesnour03: true true ahha

goomens (approved):

PR #131, opened 2026-05-07
DN-3724: Scope async jobs to a worker group

dj-petersen:
Mongo script to find the actions from the test button:
`db.jobs.find({IsSynchronous: false, SourceName: "TestDelayedJob"})`

I changed the JobStatus in the mongodb from number to string of the enum value. The migration script to update the old values to the new enum values.

```
db.jobs.updateMany({ Status: 0 }, { $set: { Status: "Pending" } })
db.jobs.updateMany({ Status: 1 }, { $set: { Status: "Completed" } })
db.jobs.updateMany({ Status: 2 }, { $set: { Status: "Failed" } })
```


UvA.Workflow/Jobs/JobRepository.cs
```diff
@@ -5,25 +5,46 @@ namespace UvA.Workflow.Jobs;
...
+        var pending = f.Eq(j => j.Status, JobStatus.Pending);
+        var expiredRunning = f.And(
+            f.Eq(j => j.Status, JobStatus.Running),
+            f.Lt(j => j.ClaimedUntil, now)
         );
-        return await result.ToListAsync(ct);
+
+        var claimable = f.Or(pending, expiredRunning);
+
+        var filter = f.And(correctGroup, hasStarted, claimable);
```
goomens: Do you really need all this composition (i.e. can the Mongo library not translate the logical expression in a single filter?)

goomens (dismissed):

UvA.Workflow/Jobs/JobRepository.cs
dj-petersen: It can but it became a unreadable mess of logic

UvA.Workflow.Api/appsettings.json
```diff
@@ -61,6 +61,9 @@
   }, 
   "Encryption": {
     "Secret": "replace-me"
+  },
+  "Worker": {
+    "WorkerGroup": "tst"
```
annesnour03: We should probaby set this to local, or not set this at all right (because of the default in `WorkerOptions`, which now isn't applied)? Now by default if connected to the tst DB, my local machine will pick up those tasks

UvA.Workflow/Jobs/Job.cs
```diff
@@ -5,6 +5,7 @@ namespace UvA.Workflow.Jobs;
 public enum JobStatus
 {
     Pending,
+    Running,
```
annesnour03: Adding this enum here 'shifts' the values up by one. So previously where completed was '1' in the DB, it now means running. So there are a lot of tasks that are 'running' now. Is this okay?

UvA.Workflow.Api/appsettings.json
dj-petersen: Good catch, i've updated it

UvA.Workflow/Jobs/Job.cs
dj-petersen: I've moved it down and changed it in the db to strings

goomens (dismissed):

goomens (approved):

PR #132, opened 2026-05-11
feature/DN-3749-BE-FE-Organisations-list

codure:
Added MongoDB Organization management (create, findById, findByName) and added those to the CreateExternalUser flow


UvA.Workflow.Persistence.Mongo/UvA.Workflow.Persistence.Mongo.csproj
```diff
@@ -12,6 +12,8 @@
         <PackageReference Include="MongoDB.Driver" Version="3.5.2" />
         <PackageReference Include="Serilog" Version="4.3.0" />
         <PackageReference Include="Microsoft.Extensions.Options" Version="10.0.1" />
+        <PackageReference Include="SharpCompress" Version="0.48.0" />
```
goomens: Why are these here?

UvA.Workflow/Organizations/InstanceOrganization.cs
```diff
@@ -0,0 +1,18 @@
+using MongoDB.Bson.Serialization.Attributes;
+
+namespace UvA.Workflow.Organizations;
+
+public class InstanceOrganization
```
goomens: Isn't `InstanceOrganization` just the same as `Organization`?

UvA.Workflow/Organizations/OrganizationService.cs
```diff
@@ -0,0 +1,48 @@
+using Microsoft.Extensions.Caching.Memory;
+
+namespace UvA.Workflow.Organizations;
+
+public class OrganizationService(IOrganizationRepository organizationRepository, IMemoryCache memoryCache)
```
goomens: Do we need this caching here? Why?

UvA.Workflow/Persistence/MongoDBIndexInitializers.cs
```diff
@@ -0,0 +1,19 @@
...
+        await OrganizationsIndexes(ct);
+    }
+
+    private async Task OrganizationsIndexes(CancellationToken ct = default)
+    {
+        var collection = database.GetCollection<Organization>("organizations");
+        var keys = Builders<Organization>.IndexKeys.Ascending(o => o.Name);
+        var options = new CreateIndexOptions { Name = "organizations_name" };
+
+        await collection.Indexes.CreateOneAsync(new CreateIndexModel<Organization>(keys, options),
```
goomens, line 16: Does this just silently ignore it if it already exists?

UvA.Workflow/WorkflowInstances/AnswerConversionService.cs
```diff
@@ -137,4 +139,24 @@ private async Task<BsonValue> ConvertUser(JsonElement value, CancellationToken c
...
+
+    private async Task<BsonValue> ConvertOrganization(JsonElement value, CancellationToken ct)
+    {
+        try
+        {
+            var instanceOrganization = value.Deserialize<InstanceOrganization>(Options);
+            if (instanceOrganization == null || string.IsNullOrWhiteSpace(instanceOrganization.Id))
+                return BsonNull.Value;
+
+            var organization = await organizationService.GetOrganization(instanceOrganization.Id, ct);
```
goomens: Why do we need this here?

UvA.Workflow.Persistence.Mongo/UvA.Workflow.Persistence.Mongo.csproj
codure: Fair question. I received moderate security warnings about a specific version but looking a bit closer I found they were only used implicitly referenced by MongoDB. I'll see if we can safely update the MongoDB drivers instead (the security issue seems to be fixed in SharpCompress 0.48.0.)

image
image

UvA.Workflow.Api/Users/UsersController.cs
```diff
@@ -32,7 +32,9 @@ public async Task<ActionResult<UserDto>> Create([FromBody] CreateUserDto dto, Ca
             UserName = dto.UserName,
             DisplayName = dto.DisplayName,
             Email = dto.Email,
-            PreferredLanguage = dto.PreferredLanguage
+            PreferredLanguage = dto.PreferredLanguage,
+            Organization = dto.Organization,
+            ProviderKey = dto.IsExternal == true ? UserProviderKeys.External : UserProviderKeys.Internal
```
annesnour03: I think we want to use the eduid keys right?

UvA.Workflow/Persistence/MongoDBIndexInitializers.cs
codure: Yes.

UvA.Workflow/Organizations/OrganizationService.cs
codure: Not really. I started this ticket with the idea of a separate 'Organization' component which was copied from how User works. It was later removed to be a part of User, but this still remained. (I've removed it from the org. service)

UvA.Workflow/Organizations/InstanceOrganization.cs
codure: (yes, it's already removed)

UvA.Workflow/WorkflowInstances/AnswerConversionService.cs
codure: I started the ticket with the idea of an Organization Input/Component but that was later removed to make it simpler.

goomens (approved):

PR #134, opened 2026-05-13
DN-3760 refactor: user email handling and search functionality

annesnour03:
https://uvanose.atlassian.net/browse/DN-3760
 
So this PR changes a few things:
1. Email validation, so the frontend can show an error if the email already exists or is internal
2. Create a separate path for creating external users from the frontend in `SaveAnswer`. We previously 'provisioned' the external users later on in the process, but that causes us to be stuck in a situation where we kind of had to hardcode that the `providerKey` was `"eduId"`, just because `IsExternal` is `true`. Now this is more decoupled with an `IExternalUserService`.
3. Cleaned up the providerkeys that were not used anymore.

This now uses https://github.com/UvA-FNWI/UvA.DataNose.WebAPI/pull/1309
> [!CAUTION]
> Revert  [59efcdd](https://github.com/UvA-FNWI/workflow-api/pull/134/commits/59efcdd5afbf07bff5fde7afbb214b36598c1679) before merging, it couples this PR env with a DN2 PR for testing.


UvA.Workflow/WorkflowInstances/AnswerConversionService.cs
```diff
@@ -119,22 +115,49 @@ private BsonValue ConvertCurrency(JsonElement value)
...
+        if (user == null)
+        {
+            if (userInput.IsExternal && !string.IsNullOrWhiteSpace(userInput.Email))
+            {
+                var existingExternalUser = await userRepository.GetByEmail(userInput.Email, ct);
+                if (existingExternalUser != null && UserProviderKeys.IsExternal(existingExternalUser.ProviderKey))
+                    user = existingExternalUser;
+            }
+
+            // External users must already exist; don't recreate them from answer payloads.
```
goomens, line 143: Why do we handle external users differently from internal ones here? (i.e. why not create the external ones in this step as well?)

UvA.Workflow/Users/RepositoryUserSearchSource.cs
```diff
@@ -0,0 +1,17 @@
+namespace UvA.Workflow.Users;
+
+public class RepositoryUserSearchSource(IUserRepository userRepository) : IUserSearchSource
```
goomens, line 3: I think maybe we still don't want this to return "internal" users that don't exist in the external source? That one has filtering by e.g. active employment and so on, otherwise once a user appears in our database it will never disappear? Maybe something for later...

(Related issue: what if someone's name or even email changes in the external source?)

goomens (dismissed):

UvA.Workflow/WorkflowInstances/AnswerConversionService.cs
annesnour03: It's because we don't know about the provider in this context/project. This is something that is present in the api layer / authentication project

UvA.Workflow/Users/RepositoryUserSearchSource.cs
annesnour03: This is a good point, we kind of have two sources of truth now, the DN users and the users we save to mongo. 

Is it an option to only search through the eduid users in `RepositoryUserSearchSource`, and leave the rest of the users to DN?

goomens: I think that makes sense yes

codure (dismissed):

codure (dismissed):

UvA.Workflow.Api/appsettings.json
```diff
@@ -39,7 +39,7 @@
   "FileKey": "bEIaRcd3spsh08QfzyiiMnOnk1C5Qy6YNA21BmjIQ72nrY3CINbEIaRcd3spsh08QfzyiiMnOnk1C5Qy6YNA21BmjIQ72nrY3CIN",
   "ImpersonationKey": "h8PbQEcmwuE4UD1G1rHPX7XB6P0V2a8QkqvnOlJMD9AiCB3j2E5csMUrAq0Sx2xQtUKh8PbQEcmwuE4UD1G1rHPX7XB6P0V2a8Qkq",
   "DataNoseApi": {
-    "BaseAddress": "https://api-v2-tst.datanose.nl/"
+    "BaseAddress": "https://api-v2-pr-1309.datanose.nl/"
```
goomens: Errr 😊

goomens (dismissed):

codure (approved):

PR #136, opened 2026-05-20
DN-3789: Creating Project-PP type on tst causes error logging

dj-petersen:


goomens (dismissed):

goomens:
@danielle-devv Haha what is happening here!? Can we fix it in such a way that we won't run into this issue again? Because Reinout had a similar problem with his workflow so I guess it will be something like this as well...

dj-petersen:
The problem was that properties were defined in two places. I've added checks for properties and globalActions. Props are merged, and only throws if a prop with the same name is added in two places. globalActions throws if it is defined in more than one file.

goomens (approved):

PR #139, opened 2026-05-27
Feature/dn 3770 jobs history

codure:
JobsController (AdminOnly) for Listing, Viewing and (Re)Running Jobs (per instance)


UvA.Workflow.Persistence.Mongo/JobRepository.cs
```diff
@@ -8,6 +8,23 @@ public class JobRepository(IMongoDatabase database, IOptions<WorkerOptions> work
...
+    public async Task<Job?> GetById(string id, CancellationToken ct)
+    {
+        var filter = Builders<Job>.Filter.Eq("_id", ObjectId.Parse(id));
+        return await _jobCollection.Find(filter).FirstOrDefaultAsync(ct);
+    }
+
+    public async Task<IReadOnlyList<Job>> GetList(string? instanceId, CancellationToken ct)
+    {
+        var filter = string.IsNullOrWhiteSpace(instanceId)
+            ? Builders<Job>.Filter.Empty
```
goomens: I don't really think we ever want to get all jobs because that's going to be huuuge

UvA.Workflow/Jobs/Job.cs
```diff
@@ -35,6 +35,8 @@ public class Job
     [BsonRepresentation(BsonType.ObjectId)]
     public string? CreatedBy { get; set; }
 
+    public string? CreatedByDisplayName { get; set; }
```
goomens: Do we want to keep this in here instead of looking it up as needed?

UvA.Workflow/Jobs/JobService.cs
```diff
@@ -85,6 +92,9 @@ public async Task RunJob(Job job, CancellationToken ct)
             throw new Exception($"User {job.CreatedBy} not found");
         }
 
+        if (string.IsNullOrWhiteSpace(job.CreatedByDisplayName))
```
goomens: This seems superfluous if we set it on creation?

UvA.Workflow/Jobs/JobService.cs
```diff
@@ -68,6 +70,11 @@ await repository.Add(new Job
         return result;
     }
 
+    public Task<Job?> GetById(string id, CancellationToken ct) => repository.GetById(id, ct);
```
goomens: Doesn't it make more sense to just call the repository directly instead of having these methods here?

UvA.Workflow.Persistence.Mongo/JobRepository.cs
codure: Fair point. I'll refactor the call to "Jobs/{instanceId}/{jobId}" as well then to make them more in line with the rest of the Api.

UvA.Workflow/Jobs/Job.cs
codure: Same idea as the 'get full list' but without that it's pretty much look'upable when needed. I'll add a lookup before returning the Dtos to have the same result.

goomens (dismissed):

goomens (approved):

PR #141, opened 2026-05-29
DN-3788 feat: add user and workflow definition APIs to include admin …

annesnour03:
…rights and instance creation permissions

https://uvanose.atlassian.net/browse/DN-3788


UvA.Workflow.Api/Users/UsersController.cs
```diff
@@ -19,7 +20,11 @@ public async Task<ActionResult<UserDto>> GetLoggedInUser(CancellationToken ct)
         if (user == null)
             return UserNotFound;
 
-        return Ok(UserDto.Create(user));
+        // Developer tooling is restricted to DataNose super admins (the SystemAdmin role). The role
+        // name arrives from DataNose via GetRolesForUser; see DataNoseDirectoryKeys.
+        var roles = await userService.GetRolesOfCurrentUser(ct);
```
codure, line 25: Might be more useful to add the full list of roles to the UserDto? That way, we can just check the list of roles in the frontend to show/hide certain stuff based on the existence. (Also useful for later checks with the upcoming "screen per role" pages)

UvA.Workflow.Api/WorkflowDefinitions/WorkflowDefinitionsController.cs
```diff
@@ -1,30 +1,41 @@
...
             .OrderBy(et => et.Index ?? int.MaxValue)
-            .ThenBy(et => et.Name);
+            .ThenBy(et => et.Name)
+            .ToArray();
 
-        return Ok(workflowDefinitionDtos);
+        var dtos = new List<WorkflowDefinitionDto>(definitions.Length);
+        foreach (var definition in definitions)
+        {
+            var canCreateInstance = await rightsService.CanAny(definition.Name, RoleAction.CreateInstance);
```
codure, line 24: Option to just add every allowed action for each workflow? Doesn't really get queried a lot, but might help with further checks/calls with for-example viewing rights or editing.

Something like:

```
    private async Task<RoleAction[]> GetAllowedActions(string workflowDefinition)
        => (await rightsService.GetAllowedActions(workflowDefinition, Enum.GetValues<RoleAction>()))
            .Select(a => a.Type)
            .Distinct()
            .ToArray();
```

codure (changes requested):

UvA.Workflow.Api/Users/UsersController.cs
annesnour03: Good idea for the screen per role, but maybe we want to wait with that. `IsSuperAdmin` already covers the only gate we need today, and computing it server-side keeps the "what counts as super admin" decision in one place rather than spreading role-name checks across the frontend. I'd rather add the full role list when the screen-per-role pages actually gets picked up and we know exactly what we need/want.

UvA.Workflow.Api/WorkflowDefinitions/WorkflowDefinitionsController.cs
annesnour03: `GetAllowedActions` only sees global roles and ignores instance scoping. So a student (global role just `[Registered]`) would show only `CreateInstance` despite being able to edit their own instance, while a coordinator would show `Execute` even where it only applies on specific steps. I think we are a level too high for this

PR #143, opened 2026-06-01
DN-3750 Handle case where an action has no form restrictions

danielleklaasen:


goomens (commented):
I'm not sure if is a good idea; it might make it too easy to assign edit rights on all forms. Maybe better to always require the explicit `<All>` there?

goomens (dismissed):

goomens (approved):

PR #152, opened 2026-06-08
DN-3812 feat: introduce Impersonator role and related actions

annesnour03:
DN-3812


UvA.Workflow.Api/WorkflowInstances/Dtos/WorkflowInstanceDto.cs
```diff
@@ -23,6 +23,7 @@ public record WorkflowInstanceDto(
     SubmissionDto[] Submissions,
     RoleAction[] Permissions,
```
goomens, line 24: I'm sort of thinking again we should consider sticking the permissions in here?

UvA.Workflow.Api/WorkflowInstances/WorkflowInstancesController.cs
```diff
@@ -90,10 +90,14 @@ public async Task<ActionResult<IEnumerable<ImpersonationRoleDto>>> GetImpersonat
         if (instance == null)
             return WorkflowInstanceNotFound;
 
-        if (!await rightsService.Can(instance, RoleAction.ViewAdminTools, RightsEvaluationMode.RealUser))
+        var currentUser = await userService.GetCurrentUser(ct);
```
goomens: Why do we have this here? Won't the controller require authorization anyway?

UvA.Workflow/Users/RightsService.cs
```diff
@@ -19,14 +19,20 @@ public class RightsService(
     IWorkflowInstanceRepository workflowInstanceRepository,
     IImpersonationContextService? impersonationContextService = null)
 {
+    // Used only for the impersonation role picker: show roles that would have
```
goomens: Can't we just use `View` here? If you can't view anything I guess there's no point trying to load the page (and then we can't forget about this if we add more actions).

goomens (dismissed):

UvA.Workflow.Api/WorkflowInstances/WorkflowInstancesController.cs
annesnour03: Oops I was trying something

UvA.Workflow.Api/WorkflowInstances/Dtos/WorkflowInstanceDto.cs
annesnour03: Permissions are evaluated in a different context right? They are computed in the "request" context, whereas the canUseAdminTools/CanImpersonate are computed against the real user. If we all do it in the permissions, the impersonate would dissapear if you impersonated a student right

goomens: Yeah okay so it would be strange to combine them in a single array? We could now also have an array per context I guess (but it's also fine like this)

goomens (approved):

PR #155, opened 2026-06-09
DN-3762 Updated Organizations for DirectoryUser services

codure:
Rewrote some "directory-based" organization parts.
The underlying issue was that there were two separate 'Organization' types. 
1. The Workflow.Organization model, which uses an ObjectId for _id identifiers.
2. the Organization model from datanose Directory calls, which used the name as both the Id and Name of the model.

I've set the Id to null and made it nullable in the model. This should help with the lookups from Directory-based organizations ,so they don't need a match in Mongodb (yet). And allows us later on to fill/update/verify these Organizations.

### DIFF
--- a/UvA.Workflow/Users/UserService.cs
+++ b/UvA.Workflow/Users/UserService.cs
@@ -75,6 +75,7 @@ public abstract class UserServiceBase(IUserRepository userRepository, IMemoryCac
         }
 
         memoryCache.Set(cacheKey, user, UserCacheExpiration);
+
         return user;
     }
 
@@ -103,6 +104,7 @@ public abstract class UserServiceBase(IUserRepository userRepository, IMemoryCac
 public class UserService(
     ICurrentUserAccessor currentUserAccessor,
     IUserRepository userRepository,
+    IOrganizationService organizationService,
     IMemoryCache cache,
     IEnumerable<IUserDirectory> userDirectories,
     IEnumerable<IUserSearchSource> userSearchSources)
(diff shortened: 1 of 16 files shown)


UvA.Workflow/Users/UserService.cs
```diff
@@ -197,9 +199,11 @@ public async Task<IEnumerable<UserSearchResult>> FindUsers(string query, bool in
         {
             foreach (var directory in _userDirectories)
             {
-                var organization = await directory.GetOrganization(uid, ct);
-                if (organization != null)
-                    return organization;
+                var directoryOrganization = await directory.GetOrganization(uid, ct);
```
annesnour03, line 202: This now returns on the first fail right? I think we want to retry until one of the directories yields

UvA.Workflow/Users/UserService.cs
```diff
@@ -197,9 +199,11 @@ public async Task<IEnumerable<UserSearchResult>> FindUsers(string query, bool in
...
+                if (directoryOrganization == null)
+                    return null;
+
+                return await organizationService.GetOrCreateOrganization(directoryOrganization.Name, ct);
```
annesnour03: Do we need to introduce a unique constraint on the organisation name, now that we create it here?

codure: Oh we could add a 'unique' index on the name in MongoDB, as an extra security?

UvA.Workflow/Users/UserService.cs
codure: Ah... yep. fixing it

UvA.Workflow/Users/UserService.cs
codure: I've added the 'option' for unique indexes but the implementation needs a bit more refinement. 
GetOrCrreateOrganization does a get before actually creating a new organization, so it's slightly mitigated at least.

goomens (approved):

PR #156, opened 2026-06-10
DN-3762 (feat) added flag for invitations on User model

codure:


UvA.Workflow/Users/User.cs
```diff
@@ -2,6 +2,13 @@
 
 namespace UvA.Workflow.Users;
 
+public enum UserInvitationState
+{
+    Required,
```
annesnour03, line 11: What does "required" represent here? Is it that an invitation still needs to be sent out, but the user has been created? If so, can we add a comment here describing this?

annesnour03 (dismissed):

UvA.Workflow/Users/User.cs
codure: yes, exactly that. I've added some comments 👍

goomens (approved):

PR #158, opened 2026-06-10
DN-3820 fix: include bilingual text with rubric grades

annesnour03:
DN-3820


Examples/Projects/Common/ValueSets/HalfGrade.yaml
```diff
@@ -2,26 +2,44 @@ values:
   - name: "10"
     value: 10
   - name: "9,5"
+    text:
+      en: "9.5"
```
goomens, line 6: Ah I missed this ticket but maybe we should discuss tomorrow if we want this... in the CRA we always use commas to match the Dutch grading scheme (but not sure if that makes sense)

goomens (approved):

goomens (dismissed):

goomens (approved):

PR #160, opened 2026-06-11
fix: bug when creating instances through API

annesnour03:
When creating an instance, `userProperty` (e.g. Student) was always set to the current user. That's fine for self-service, but when an API creates an instance on someone's behalf (DN2 provisioning a Project for a student) the supplied student got overwritten by the API user (`__ApiUser`).

Now we only default it to the creator when the property wasn't already supplied in the initial properties.


goomens (approved):
Huh why wasn't this broken before?

annesnour03:
> Huh why wasn't this broken before?

I think it worked when we init'd the test environment, but then made changes to the create controller I guess. After that we didn't use it until the ACC init, where I had to fix something similar...

PR #161, opened 2026-06-11
DN-3805 DN-3801 DN-3835: Final grade calculation

dj-petersen:


annesnour03 (commented):
Maybe we could take some insipration how we do calculations in the CBT in DN2 here? I feel like we could benefit from this, as we currently have like a two-level structure (parts -> sources -> pages/questions). This makes it kind of difficult to extend, as we keep on adding more fixed layers.

Maybe something like this:
```cs
  public class AssessmentNode : INamed
  {
      public string Name { get; set; } = null!;
      public BilingualString? Title { get; set; }
      public decimal Weight { get; set; } = 1;

      public string? Source { get; set; }
      public List<AssessmentNode> Children { get; set; } = [];
  }
```

A node with `Source` would resolve to a submitted form/source result, while a node with Children would aggregate its children by weight. That would also avoid using special cases like “final grade vs part average vs source average”; they become the same calculation at different depths.

Do we expect arbitrary-depth assessments here, or is the current parts -> sources shape intentionally the only supported variant for now?

UvA.Workflow/WorkflowModel/Question.cs
```diff
@@ -191,6 +191,27 @@ public class PropertyDefinition : INamed
     public List<RubricEntry>? Rubric { get; set; }
 }
 
+public enum CalculationType
+{
+    Normal,
+    Add,
+    Subtract
```
codure: Is the separation of Add and Subtract necessary? This would prevent options that could be either +0.4 or -0.2 for example.

Examples/Projects/PP/Project-PP/Entity.yaml
```diff
@@ -1,10 +1,6 @@
 name: Project-PP
-title: 
-  en: Project (PP)
-  nl: Project (PP)
-titlePlural: 
-  en: Projects (PP)
-  nl: Projecten (PP)
+title: Thesis PPLE
+titlePlural: 3803THQPVY
```
codure, line 3: debug id?

Examples/Projects/PP/Assessment-PP/Entity.yaml
```diff
@@ -93,12 +109,19 @@ properties:
           - The description of the methods is not clear enough to make the research repeatable
         grades: ["5", "4,5", "4", "3", "2", "1"]
 
+  - name: MethodsComments
+    type: String
+    text: Comments on methods
```
codure, line 114: Is the pilot 'english only'? I've noticed a lot of fields with only the english translation.

codure (changes requested):

Examples/Projects/PP/Assessment-PP/Entity.yaml
goomens: Yes. these courses are English only

Examples/Projects/PP/Project-PP/Entity.yaml
goomens: This is the actual course code for this course which I put in this field for now 😊

UvA.Workflow/WorkflowModel/Question.cs
goomens: Mmm yes good point. I'll change it to just `Add` and allow negative values

codure (approved):

PR #164, opened 2026-06-13
Fix/dn 3796

annesnour03:
DN-3796

I'm not 100% sure about this approach, it feels somewhat weird with the date checking, but I couldn't get it working with only a `WasFormEverSubmitted`. It would also count in previous form submissions, so if you had once submitted a form, and then in a new unsubmitted variant because that would also be seen as 'this form once was submitted' -> file wouldn't be deleted here...


goomens:
I think your approach makes sense but it is not entirely consistent with the logging that we already had in place. So we may want to change that to match this, but for now I have aligned the deletion logic with that (since I don't really dare touch this right now...)

While doing that I found out that the versioning was entirely broken for files. We had the following behaviour:
- Submit version 1 without test document
- Reject
- Add a test document in version 2
- The new test document also appears in version 1 (even before submitting)
- Remove the test document from version 2
- The test document is still there in version 1

(Uploading a document skipped versioning entirely)

This should now be fixed...

codure (dismissed):

codure (approved):

PR #177, opened 2026-06-19
fix/DN-3846-sequential-assessment-steps

codure:
Made steps start and ProposalAssessment sequential.

### DIFF
--- a/Examples/Projects/Project-RMSS/Steps/ProposalAssessmentReviewer.yaml
+++ b/Examples/Projects/Project-RMSS/Steps/ProposalAssessmentReviewer.yaml
@@ -4,7 +4,12 @@ title:
   nl: Beoordeling (voorstelbeoordelaar)
 
 condition:
-  event: Start
+  logical:
+    operator: Or
+    children:
+      - event: Start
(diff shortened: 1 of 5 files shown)


Examples/Projects/Project-RMSS/Steps/ProposalAssessmentReviewer.yaml
```diff
@@ -4,7 +4,12 @@ title:
   nl: Beoordeling (voorstelbeoordelaar)
 
 condition:
-  event: Start
+  logical:
```
goomens: Is this condition still needed now?

codure: Yes, this makes it so that when a reviewer submits their assessment, they at least see their own response. otherwise they either get just the Start form, or an empty 'Nothing has been submitted' step. (I could be wrong, it has been a hot day)

codure: They were not necessary anymore! 👍

goomens (approved):

PR #187, opened 2026-06-26
feature/DN-3759

codure:
Jira: [DN-3759](https://uvanose.atlassian.net/browse/DN-3759)

[DN-3759]: https://uvanose.atlassian.net/browse/DN-3759?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Examples/Projects/Context/Entity.yaml
```diff
@@ -22,4 +22,10 @@ properties:
     type: "[User]!"
 
   - name: StudyManual
-    type: String!
\ No newline at end of file
+    type: String!
+    
+  - name: ConfidentialAdviser
```
goomens: Adviser or Advisor? 😊

UvA.Workflow.Api/Users/UsersController.cs
```diff
@@ -87,6 +87,42 @@ public async Task<ActionResult<UserDto>> GetById(string id, CancellationToken ct
         return Ok(UserDto.Create(user));
     }
 
+    [HttpPut("{id}/email")]
+    public async Task<ActionResult<UserDto>> UpdateEmail(
+        string id,
+        [FromBody] UpdateUserEmailDto dto,
+        CancellationToken ct)
+    {
+        await rightsService.EnsureAuthorizedForAction(RoleAction.Edit);
```
goomens: I don't think this works; now you can either edit all users or none. Maybe this should take an instance ID and allow editing the email if you are authorized to edit that field in the instance?

Additionally, I guess it should also actually edit the instance data and not just the user record? Because we store the email in both places, I guess? (and that way we will also have audit logging and some of the existing SaveAnswer logic can be re-used)

UvA.Workflow.Api/Users/UsersController.cs
```diff
@@ -95,7 +131,7 @@ public async Task<ActionResult<IEnumerable<UserSearchResultDto>>> Find(string qu
         return Ok(searchResults.Select(UserSearchResultDto.Create));
     }
 
-    private async Task<ObjectResult?> ValidateEmail(string email, CancellationToken ct)
+    private async Task<ObjectResult?> ValidateEmail(string email, CancellationToken ct, string? ignoredUserId = null)
```
goomens: If you do the equality check first I think maybe you don't need the `ignoredUserId` here?

Examples/Projects/Context/Entity.yaml
codure: Ye olde "en vs us". I'll vote for Advisor.

UvA.Workflow.Api/Users/UsersController.cs
codure: Check! Had a feeling this wasn't complete... It should definitely have the instance data updated as well. I'll make it a requirement as well as it should stay in sync, even if only visually.

UvA.Workflow/WorkflowInstances/EffectService.cs
```diff
@@ -184,12 +185,46 @@ private async Task EnsureExternalAccounts(
...
                 EduIdInviteDeliveryMode.SendEmail,
                 ct);
+            if (result.User != null)
+                updatedExternalUsers[email] = InstanceUser.FromUser(result.User);
         }
+
+        if (updatedExternalUsers.Count == 0)
+            return;
+
+        UpdateInstanceUserProperties(instance, property, rawValue, updatedExternalUsers);
```
goomens, line 200: What exactly needs to be updated here?

goomens (dismissed):

UvA.Workflow/WorkflowInstances/EffectService.cs
codure: The current 'invite a person' flow stores the user data in both the instance and as a separate user document.
The RelatedUsers uses the instance version of this user to check the invitation state as a flag to see if it can still edit the email address. (atm, only before sending an actual invite to the email address)

This moment is when the user actually gets invited and then updated for both the user and instance.

In hindsight, might've been easier to just do add a quick user lookup instead of this.

goomens (approved):

PR #189, opened 2026-06-29
DN-3804: Final grade rounding

dj-petersen:


Examples/Projects/PP/Project-PP/Entity.yaml
```diff
@@ -28,6 +28,8 @@ properties:
     type: Assessment-PP!
     
 assessments:
+  gradingBasis: decimal
```
goomens: I think ultimately we want to be able to retrieve this from the linked course context (see the first acceptance criterium in the ticket) because it comes from an external system. Can we use some kind of reference expression here?

UvA.Workflow.Api/Assessments/Dtos/AssessmentDto.cs
```diff
@@ -106,11 +107,25 @@ public AssessmentDto Create(
...
-            : (decimal?)null;
+        if (assessmentConfig == null) return new AssessmentDto(id, parts.ToArray(), null, null);
 
-        return new(id, parts.ToArray(), RoundToTwo(finalGrade ?? 0));
+        bool isGradingComplete = assessmentConfig.Parts
+            .All(p => p.Sources.All(s => contextList.Any(c => c.Form.Name == s.Name)));
+
+        if (!isGradingComplete)
+            return new AssessmentDto(id, parts.ToArray(), null,
+                new BilingualString("Grading not yet complete", "Beoordeling nog niet volledig"));
```
goomens: Doesn't this kind of belong in the frontend?

UvA.Workflow/Assessments/AssessmentService.cs
```diff
@@ -5,7 +5,7 @@ namespace UvA.Workflow.Assessments;
 
 public static class AssessmentService
 {
-    public static decimal CalculateFinalGrade(AssessmentConfiguration config,
+    public static float CalculateFinalGrade(AssessmentConfiguration config,
```
goomens: Why float instead of decimal?

UvA.Workflow.Api/Assessments/Dtos/AssessmentDto.cs
```diff
@@ -106,11 +107,25 @@ public AssessmentDto Create(
...
+
+        var calculatedFinalGrade = AssessmentService.CalculateFinalGrade(assessmentConfig, domainPartResults);
```
goomens: Ah this was not in the ticket, but now I think about it: it's probably good to return both the rounded and the unrounded grade to the frontend. In the CBT we have a little mouseover for rounded grades showing the original

UvA.Workflow/Assessments/AssessmentService.cs
dj-petersen: Since the grade is rounded it didn't feel useful to store it in 16 bytes (decimal) instead of 4 (float)

Examples/Projects/PP/Project-PP/Entity.yaml
dj-petersen: I've created an Enrich function to get it from the context

UvA.Workflow.Api/Assessments/Dtos/AssessmentDto.cs
dj-petersen: I removed the text but kept the null when there is no final grade yet

UvA.Workflow.Api/Assessments/Dtos/AssessmentDto.cs
dj-petersen: I've added it

UvA.Workflow/WorkflowModel/AssessmentConfiguration.cs
```diff
@@ -29,6 +29,25 @@ public class AssessmentConfiguration
     /// Determines whether grades between 5 and 6 are allowed
     /// </summary>
     public bool? GradeGap { get; set; }
+
+    public AssessmentConfiguration Enrich(ObjectContext context)
+    {
+        var contextGradingBasis = context.Get("Course.GradingBasis") is string s
```
goomens, line 35: This is super specific but maybe this entire functionality is super specific...

goomens (approved):

PR #191, opened 2026-06-29
Fix/dn 3864 parallel step conditions

codure:
"Bugfix" with a rabbit hole of fixes and updates. (and an unrelated RMSS Assessment step icon)

- Step versions are now based on event history instead of only live/current instance state.
- Parallel child steps now produce versions correctly when all required child conditions complete.


UvA.Workflow.Api/WorkflowInstances/Dtos/WorkflowInstanceDtoFactory.cs
```diff
@@ -181,7 +185,7 @@ private StepVersionDto CreateStepVersionDto(StepVersion stepVersion, WorkflowIns
                     continue;
                 }
 
-                if (!allowedForms.Contains(form.Name))
+                if (!allowedViewActions.Any(action => action.MatchesForm(form.Name)))
```
annesnour03, line 221: We changed historical form visibility from `allowedForms.Contains(form.Name)` to `MatchesForm(...)`, but those don’t behave the same for wildcard View actions. `AllForms` returns the literal `"<All>"` string, so a `{ type: View, form: <All> }` role that previously hid old-version submissions would now show them.

No current role uses `<All>` for View, so this doesn’t break anything today, but it does mean historical submissions would become visible for anyone who adds that role later. Was that intentional?

UvA.Workflow/Versioning/StepVersionService.cs
```diff
@@ -226,4 +219,59 @@ private static List<StepVersion> BuildStepVersions(
...
+                    Children = step.Children
+                        .Select(GetCompletionCondition)
+                        .Where(condition => condition != null)
+                        .Cast<Condition>()
+                        .ToArray()
+                }
+            };
+    }
+
+    private static bool IsEventConditionMet(Condition? condition, IEnumerable<string> eventIds)
```
annesnour03: We already have `Condition.IsMet(context)`, do we need a second condition evaluator that only understands events?

UvA.Workflow/Submissions/AnswerService.cs
```diff
@@ -61,7 +61,10 @@ private async Task SaveAndLogAnswer(QuestionContext context, BsonValue? currentA
             {
                 // if the journal entry was replaced, the old value isn't stored either
                 isReplaced = await instanceJournalService.LogPropertyChange(instance.Id,
-                    PropertyChangeEntry.Create(context.PropertyDefinition, currentAnswer, user), ct);
+                    PropertyChangeEntry.Create(
```
annesnour03: This change might have some impact right? Journal entries already written for nested questions still have the old leaf-only path

UvA.Workflow/WorkflowInstances/WorkflowInstanceService.cs
```diff
@@ -84,7 +86,12 @@ public async Task<WorkflowInstance> GetAsOfVersion(string instanceId, int versio
     public async Task<WorkflowInstance> GetAsOfTimestamp(string instanceId, DateTime timestamp, CancellationToken ct)
     {
         var version = await GetVersionAtTimestamp(instanceId, timestamp, ct);
-        return await GetAsOfVersion(instanceId, version, ct);
+        var instance = await GetAsOfVersion(instanceId, version, ct);
+
+        instance.Events = await RebuildEventsUntil(instanceId, timestamp, ct);
+        RecalculateCurrentStep(instance);
```
annesnour03: I feel like this does a lot now. Could historical reconstruction live in a dedicated snapshot service? It could load the
instance, journal, and event history once, then reconstruct properties, events, and CurrentStep
consistently. We could have like a replay and resolve function, now its a bit scattered and hard to follow.

This might also help with performance, as we are kind of double reading here, for each version

annesnour03 (commented):
Nice work on this. This part of the app is pretty complex, so kudos for digging into it. I left a few comments/questions around behavior and structure.

UvA.Workflow/Versioning/StepVersionService.cs
codure: I'm refactoring it so that we can call Condition.IsMet(eventIds), still only applies to Logical and Event conditions though.

The IsMet(context) evaluation validates against live values and dates and such, which the 'verions' don't really have to do again, as the events have already been determined 'completed' at that time. 
We 'can' rebuild the context 'per event' to replay the conditions again for a given time of each event, but I think this will become quite expensive.

UvA.Workflow/Submissions/AnswerService.cs
codure: Yes, this one will probably need a migration script. I'm testing and evaluating this part atm.

UvA.Workflow.Api/WorkflowInstances/Dtos/WorkflowInstanceDtoFactory.cs
codure: Nope! wasn't intentional. It does sound like the intended way of using the "all" flag though. Are there situations that will have a role with "All" but you don't want to show specific forms? I can see either option being a valid choice

UvA.Workflow/WorkflowInstances/WorkflowInstanceService.cs
codure: Agreed. I'm refactoring it to use a (Journey, Events) input which will be reused for every step/version.

codure:
I've updated the WorkflowInstanceFactoryDto so that it uses instance-wide history instead of fetching/reconstructing per step.
Still needs a small migration for journal entries.. (Working on it.)

UvA.Workflow/Submissions/AnswerService.cs
codure: This will 'quietly' phase out by itself. If needed, we can always track specific cases and fix them manually.

UvA.Workflow/WorkflowInstances/WorkflowInstanceService.cs
```diff
@@ -84,7 +90,45 @@ public async Task<WorkflowInstance> GetAsOfVersion(string instanceId, int versio
...
+        WorkflowInstance instance,
+        DateTime timestamp,
+        WorkflowInstanceHistory history)
+    {
+        var instanceAtTimestamp = CloneInstance(instance);
+        var version = GetVersionAtTimestamp(history.Journal, timestamp);
+
+        if (history.Journal != null)
+        {
+            // Revert all changes after the target version against the cloned live instance.
```
annesnour03: This looks oddly similar to the logic in `GetAsOfVersion`, can we maybe share some of this logic?

UvA.Workflow/WorkflowInstances/WorkflowInstanceService.cs
```diff
@@ -84,7 +90,45 @@ public async Task<WorkflowInstance> GetAsOfVersion(string instanceId, int versio
     public async Task<WorkflowInstance> GetAsOfTimestamp(string instanceId, DateTime timestamp, CancellationToken ct)
```
annesnour03: This now is not used right? Can we remove it? I think the `GetVersionAtTimestamp` also is dead then, aswell as `RebuildEventsUntil` and `GetEventLogEntriesForInstanceUntil`. I think we should do something with this

UvA.Workflow/Versioning/StepVersionService.cs
annesnour03: Okay looks good, I see its an extension method now. It goes through the parts, but silently returns `false` for Value/Date/Deadline parts.  I think this is fine as a deliberate limitation, but a mixed `Ends: And(event, value)` would silently never version. Can we maybe throw something of an unsupported part error?

UvA.Workflow.Api/WorkflowInstances/Dtos/WorkflowInstanceDtoFactory.cs
annesnour03: Let's keep `MatchesForm`. It just makes history consistent with the live view, an `<All>` View role can already see every current submission, so seeing them in the version history is expected. If we ever need "can view live but not history" that'd be a separate thing

annesnour03 (changes requested):
Few small things, I think we're almost there 😄 .

UvA.Workflow/WorkflowInstances/WorkflowInstanceService.cs
codure: Agreed and done. I've removed the unused functions from the service.

UvA.Workflow/WorkflowInstances/WorkflowInstanceService.cs
codure: I've added a private enrich step that both use now.

UvA.Workflow/WorkflowInstances/WorkflowInstanceService.cs
annesnour03: > Agreed and done. I've removed the unused functions from the service.

Yes nice, can we also remove `GetEventLogEntriesForInstanceUntil` now?

annesnour03 (dismissed):

codure:
I've fixed the merge conflicts, requires a new approve though

annesnour03 (dismissed):

UvA.Workflow.Api/WorkflowInstances/Dtos/WorkflowInstanceDtoFactory.cs
```diff
@@ -133,49 +134,63 @@ private async Task<Dictionary<string, List<StepVersion>>> GetStepVersionsMap(
...
+        WorkflowInstanceHistory instanceHistory,
         ObjectContext context,
-        string[] allowedForms)
+        CancellationToken ct)
     {
         var workflowDef = modelService.WorkflowDefinitions[instance.WorkflowDefinition];
         var versions = stepVersionsMap.GetValueOrDefault(step.Name);
 
+        if (versions is { Count: > 0 } && step.HasEnded(context))
+            versions.RemoveAt(versions.Count - 1);
```
annesnour03: This mutates inplace and assumes  the list is sorted ascending by `SubmittedAt` I guess? Can we make this more clear?

```c#
 // The newest version is the step's live state, already rendered as the current submission.
var versions = stepVersionsMap.GetValueOrDefault(step.Name)
            ?.OrderByDescending(version => version.SubmittedAt)
            .Skip(step.HasEnded(context) ? 1 : 0);
```

UvA.Workflow/Versioning/StepVersionService.cs
```diff
@@ -15,6 +15,11 @@ public record StepVersion
 public interface IStepVersionService
 {
     Task<List<StepVersion>> GetStepVersions(WorkflowInstance instance, string stepName, CancellationToken ct);
+
```
annesnour03: Do we need the previous version of `GetStepVersions`? We now overload it, but I'm not sure the old one is used. I also think it returns something different for the same stepname, so that might be confusing.

The (old) async overload seems to return the versions of the parent when you ask for a child, and the new one returns the one of a child. Can we remove the old one so we just have one truth (inc controller, the frontend doesn't even seem to use it)?

codure: I've checked the endpoint and it is indeed not in use. I'll remove it. 
(there are a lot of unit tests tho so removing it will flood this PR a bit)

codure: Yep. Done aaaand done

annesnour03 (dismissed):

goomens (approved):

PR #192, opened 2026-06-30
DN-3871 fix: Titles are not shown in development tab for some project types

dj-petersen:


goomens (dismissed):
Will this break existing instances? Do we need some kind of migration?

Edit: oh I guess not because that WorkingTitle property wasn't actually used?

UvA.Workflow/WorkflowModel/WorkflowDefinition.cs
```diff
@@ -45,7 +45,7 @@ public class WorkflowDefinition : INamed
     /// </summary>
     public string? InstanceTitle { get; set; }
 
-    public Template? InstanceTitleTemplate => field ??= Template.Create(InstanceTitle);
+    public Template? InstanceTitleTemplate => field ??= Template.Create(InstanceTitle) ?? new Template("{{ Title }}");
```
goomens: Isn't this a bit too magic?

dj-petersen: Yes maybe it is, I will try a different route and inherit it from the base project. This can still cause titles to be left empty when the project does not inherit from the base project, but it is more generic for the system.

goomens (approved):

PR #194, opened 2026-07-01
DN-3776: Switch Graph mail to client/secret app credentials

dj-petersen:
I'm still testing the deployment settings


Deployment/workflow-api-pr/values.yaml
```diff
@@ -18,8 +18,8 @@ datanoseApi:
 graph:
   tenantId: tenantId
   clientId: clientId
-  userAccount: datanose@amsuni.onmicrosoft.com
-  senderId: 0c6948d0-009a-4796-9d94-e0b56bc3ea9c
+  clientSecret: clientSecret
```
goomens: This is a secret so it doesn't need to be here

Deployment/workflow-api-pr/values.yaml
```diff
@@ -98,3 +98,7 @@ azure:
       key: Services__DataNose__Token
     - name: dn2-api-key
       key: Authentication__ApiKey
+    - name: workflow-graph-client-id
+      key: GraphMail__ClientId
```
goomens: This isn't a secret I think?

Deployment/workflow-api/values.yaml
```diff
@@ -17,8 +17,8 @@ datanoseApi:
 graph:
   tenantId: tenantId
   clientId: clientId
-  userAccount: datanose@amsuni.onmicrosoft.com
-  senderId: 0c6948d0-009a-4796-9d94-e0b56bc3ea9c
+  clientSecret: clientSecret
+  senderId: no-reply-milestones@uva.nl
```
goomens, line 20: Maybe this shouldn't be the default for tst/acc?

dj-petersen: For now we have one mailbox. Do we want additional mailboxes for test and acc? Or keep sending from testen-dn-fnwi@uva.nl?

To distinguish I did add a prefix to the email's subject with the worker it was send from (except for production of course).

goomens: Yes, the prefix is nice! For acc and tst we want to keep sending to and from testen-dn-fnwi I think, let's keep the 'real' address just for production

Deployment/workflow-api-pr/values.yaml
dj-petersen: This is the Entra AppId which will be different for test, acc and prod. So don't we need it to be in the keyvault?

goomens: No, because it is not secret. It can be configured a value in the GitOps repo

goomens (approved):

PR #199, opened 2026-07-06
Feature/dn 3863

annesnour03:
DN-3863

This PR introduces/changes a few things. Firstly, we are now able to send mails to
multiple recipients at once. This is true for all the email recipient fields (`to`,
`cc` and `bcc`), and a single email is sent to the whole set.

Because of this, we also now have some extra logic to decide in what language we should
send the mails. Currently we send it in `nl` if everyone is `nl`, otherwise it's `en`.

We previously had a `toAddress` field, which was used to send to a raw email address. This was for example
a fixed one like `office@uva.nl`, or one built from a `{{ }}` template, rather than to
a workflow user. It existed because the old `to` could only point at a single user
property. That distinction is gone: `to` now classifies each entry itself (user lookup
vs. address vs. template), so `toAddress` is redundant and has been removed.

### What `to` (and `cc`/`bcc`) now supports
| Capability | Example | Resolves to |
|---|---|---|
| User / `[User]` lookups | `to: Supervisor` | the user, or **every** user in a `[User]` list |
| Literal addresses | `to: office@uva.nl` | that address |
| Templated addresses | `to: "{{ ContactEmail }}"` | the template evaluated as an address (dropped if it evaluates to blank) |
| Mixed recipient lists | `to: [Supervisor, office@uva.nl, "{{ ContactEmail }}"]` | all of the above in one field, deduplicated by address |
| Template defaults & overrides | `template: X` + `to: Supervisor` | inline `to` overrides the mail template's `to`; otherwise the template's default is used |
| Nested lookup enrichment | `to: Course.Coordinator` | referenced properties (`Course`) are enriched from the DB before the `[User]` behind them is read |

An entry is treated as an address when it contains `@` or a `{{ }}` expression; anything
else is a property/role lookup.


UvA.Workflow/Notifications/MailBuilder.cs
```diff
@@ -8,27 +10,36 @@ public class MailBuilder(
     IMailLayoutResolver layoutResolver,
     IConfiguration configuration,
     IOptions<WorkerOptions> workerOptions,
-    IHostEnvironment environment)
+    IHostEnvironment environment,
+    ILogger<MailBuilder> logger)
 {
-    public async Task<MailMessage> BuildAsync(
+    public Task<MailMessage> BuildAsync(
```
goomens: Does this still need to be a task?

UvA.Workflow/Notifications/MailBuilder.cs
```diff
@@ -8,27 +10,36 @@ public class MailBuilder(
...
         SendMessage sendMail,
         ModelService modelService,
-        CancellationToken ct = default)
+        CancellationToken ct = default,
+        ObjectContext? context = null)
     {
         var resolvedMail =
             ResolveMessageContents(modelService.WorkflowDefinitions[instance.WorkflowDefinition], sendMail);
-        var context = modelService.CreateContext(instance);
+        context ??= modelService.CreateContext(instance);
```
goomens: When is context null? Because then it will not be enriched so might not quite work

goomens (dismissed):

UvA.Workflow/Notifications/MailBuilder.cs
annesnour03: Doesn't need to be, we can change that!

UvA.Workflow/Notifications/MailBuilder.cs
annesnour03: Well it was more to get around unit tests, but the normal code path always goes through `InstanceService.BuildMail`, so `null` is not really a valid option. I'll make it required

goomens (approved):

PR #201, opened 2026-07-07
DN-3886 feat: harmonise the project workflows onto a common base

annesnour03:
DN-3886

The workflows kept re-declaring the same properties, roles and boilerplate. This adds a shared Project-Base with that common vocabulary, so Project, Community Project, PA and RMSS inherit it and only keep what's specific to them. There's also a small parser change so definitions can inherit in layers rather than a single level.

Also cleaned up some RMSS parts that were not used anymore.


goomens (dismissed):

UvA.Workflow.Tests/Fixtures/Project-RMSS/Entity.yaml
```diff
@@ -4,7 +4,22 @@ titlePlural:
   en: Projects (RMSS)
   nl: Projecten (RMSS)
 instanceTitle: '{{ Title }}'
-inheritsFrom: Project
+inheritsFrom: Project-Base
+
+fields:
```
goomens, line 9: Shouldn't these be in base as well? (also why is `TurnitinId` here?)

(Ok, I see this is not the case in the config repo so maybe I should ignore this)

UvA.Workflow/WorkflowInstances/WorkflowInstance.cs
```diff
@@ -80,12 +80,29 @@ public InstanceEvent RecordEvent(string eventId, DateTime? date = null)
...
     }
 
+    /// <summary>
+    /// A strictly increasing UTC timestamp for a new event. Suppression orders events by a strict
+    /// comparison on <see cref="InstanceEvent.Date"/>, so two events in one instance must never share
+    /// a timestamp; MongoDB stores millisecond precision, hence the 1ms floor.
+    /// </summary>
+    private DateTime NextEventDate()
+    {
+        var now = DateTime.UtcNow;
```
goomens, line 110: UtcNow or Now?

UvA.Workflow/WorkflowModel/ModelParser.ResetParentStep.cs
```diff
@@ -0,0 +1,226 @@
...
+            throw Error(def, "target parent step is not sequential", resetEvent, declaringNames, p.Name);
+        if (p.Children.Length < 2)
+            throw Error(def, "target parent step has fewer than two children", resetEvent, declaringNames, p.Name);
+        foreach (var d in declarings)
+            if (p.Children[0].Name == d.Name)
+                throw Error(def, "reset event is declared on the target's first child", resetEvent, d.Name, p.Name);
+        if (p.Ends != null)
+            throw Error(def, "target parent step has its own ends condition", resetEvent, declaringNames, p.Name);
+        if (p.Children[0].Children.Any())
+            throw Error(def, $"first child '{p.Children[0].Name}' of the target is not a leaf step",
```
goomens, line 140: Can we clarify why this is needed?

UvA.Workflow/WorkflowModel/ModelParser.ResetParentStep.cs
```diff
@@ -0,0 +1,226 @@
...
+
+            ValidateTarget(def, p, members.Select(m => m.Declaring).ToArray(), resetEvents[0], subtree, leaves);
+
+            var restartEvents = FirstChildRestartEvents(def, p, p.Children[0], resetEvents[0]);
+            var completionEvents = leaves
+                .SelectMany(leaf => PositiveEvents(leaf.Ends!,
+                    $"ends of leaf '{leaf.Name}' under target '{p.Name}' in workflow '{def.Name}'"))
+                .Distinct()
+                .ToArray();
+            var nestedResets = resets
```
goomens: Can you clarify what this is for?

UvA.Workflow/WorkflowModel/ModelParser.ResetParentStep.cs
```diff
@@ -0,0 +1,226 @@
...
+            foreach (var e in def.Events.Where(e => e.Suppresses != null))
+                if (e.Suppresses!.Any(restartEvents.Contains) && !resetEvents.Contains(e.Name))
+                    throw Error(def,
+                        "event explicitly suppresses a restart event of a reset target but is not a declared reset event",
+                        resetEvent: e.Name, targetStep: p.Name);
+
+            foreach (var reset in resetEvents)
+                Accumulate(generated, reset, restartEvents.Concat(nestedResets.Where(n => n != reset)));
+            foreach (var restart in restartEvents)
+                Accumulate(generated, restart, completionEvents.Where(c => c != restart).Concat(nestedResets));
```
goomens: I don't think I quite understand this; why does the _restart_ event reset all completion events and not the reset?

UvA.Workflow/WorkflowInstances/WorkflowInstance.cs
annesnour03: Now actually is more in line with what we have, I'll change it

UvA.Workflow/WorkflowModel/ModelParser.ResetParentStep.cs
annesnour03: The mechanism hangs on one set of restart events derived from the first child's ends (`FirstChildRestartEvents`). A first child with children has no single completion event, its completion is the combination of its own leaves, which suppression cannot express in one go.

UvA.Workflow/WorkflowModel/ModelParser.ResetParentStep.cs
annesnour03: Okay so there is a split between these two:
- `reset`: suppresses the restart event, plus any reset markers nested inside it (not itself)
- `restart`: suppresses every other completion event in the subtree, plus all reset markers in it (including its own)

So for the RMSS pilot for example
`reset` generates suppresion for:
ProposalReturnedForRevision  →  Start

and `restart` does every other completion event:
Start                        →  ProposalApprovedSupervisor 
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;  ProposalRejectedSupervisor
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;                                      ProposalApprovedReviewer
                   &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;                   ProposalRejectedReviewer
                   &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;                   ProposalReturnedForRevision


The reason they're split: "changes needed" only makes the proposal stop counting as submitted, so the student lands back on the form. The two assessments still count at that point, since that's the feedback the student needs in order to fix it. Resubmitting is what clears them, and it has to, because a form stays locked while its outcome still counts, so otherwise the supervisor could never assess version 2.

Maybe the naming here could be clearer too, as "restart" and "reset" basically mean the same thing in plain english. Renaming to `resubmissionEvents` and `subtreeResets`, or something else, might make this more clear.

Also worth noting the restart event isn't something you declare anywhere, it's just the first child's ends event that the generator picks up, which probably adds to the confusion. So for the RMSS pilot it go up one level, to `ProposalPhase` and go to its first child, `Start` and get the ends event from there.

UvA.Workflow/WorkflowModel/ModelParser.ResetParentStep.cs
annesnour03: Okay I've renamed this, as it was a bit misleading.

This list is here to "undo" the resets users emitted earlier. A reset event is just a normal event, so once it's fired it keeps its date and would otherwise keep counting as active for the rest of the instance's life, and any condition referencing it would "believe" that.

For RMSS, its value is:
`subtreeResets = [ProposalReturnedForRevision]`

The timeline is:

  1. `ProposalReturnedForRevision` suppresses `Start`, reopening the proposal
     form.
  2. Existing assessments remain visible while the student revises.
  3. When the student resubmits, the new Start event suppresses:
      - the old assessment outcomes
      - `ProposalReturnedForRevision` from `subtreeResets`.

That last part retires the stale “changes needed” marker so the new assessment round starts cleanly.
It also handles nested workflows, restarting an outer subtree clears reset markers belonging to inner subtrees.

goomens (dismissed):

goomens (approved):

PR #204, opened 2026-07-13
DN-3877 feature: grade submission

goomens:
Note: this includes a slight refactor of the assessment logic because it was needed outside the API project and because it needed to work in expressions.

<https://uvanose.atlassian.net/browse/DN-3877>


Examples/Projects/PP/Project-PP/Steps/FinalGrade.yaml
```diff
@@ -6,13 +6,31 @@ icon:
...
+          service: DataNose
+          operation: FinalGradeSubmission
+          inputs:
+            CourseExternalId: Course.ExternalId
+            Student: Student
+            Grade: Assessment.FinalGrade
+            ExamType: =Exam
+            ExamDate: =2026-06-17
+            Title: Title
+            Supervisor: Supervisor
```
dj-petersen: Currently the fields `CourseExternalId<-Course.ExternalId, Supervisor<-Supervisor, Examiner<-Supervisor` are missing in the Project-PP workflow which causes the service call to fail.

Examples/Projects/PP/Project-PP/Steps/FinalGrade.yaml
```diff
@@ -6,13 +6,31 @@ icon:
...
+            Examiner: Supervisor
+    
 
 ends:
   event: FinalGrade
```
dj-petersen: I don't think this is the correct order. Right now the Push to SIS action is only visible before the Coordinator has submitted the final grade.

Examples/Projects/PP/Project-PP/Steps/FinalGrade.yaml
```diff
@@ -6,13 +6,31 @@ icon:
...
     type: Submit
     roles: [Coordinator]
     form: FinalGrade
     label:
       en: Publish Grade
       nl: Cijfer publiceren
+  - name: SubmitGrade
+    type: Execute
+    label: Push to SIS
+    roles: [SystemAdmin]
```
dj-petersen: Is it the SystemAdmin that needs to do this? Also, now the SystemAdmin can push to SIS but not see the overview.

UvA.Workflow.Api/WorkflowInstances/WorkflowInstancesController.cs
```diff
@@ -129,6 +129,28 @@ public async Task<ActionResult<StartImpersonationResultDto>> StartImpersonation(
         ));
     }
 
+    [Authorize(AuthenticationSchemes = WorkflowAuthenticationDefaults.AnyScheme)]
+    [HttpGet("instances/{workflowDefinition}/full")]
```
dj-petersen, line 137: Is this endpoint for the new screens? Do we use it now in the front end?

UvA.Workflow/Assessments/AssessmentService.cs
```diff
@@ -3,191 +3,105 @@
...
+                PartConfig = partConfig
+            };
+            partResults.Add(result);
         }
 
-        return totalSourceWeight == 0
-            ? 0
-            : weightedSum / totalSourceWeight;
-    }
+        if (assessmentConfig == null) return new AssessmentResult { PartResults = partResults };
```
dj-petersen: Shouldn't this clause be at the top of this function?

UvA.Workflow/Assessments/AssessmentService.cs
```diff
@@ -3,191 +3,105 @@
...
+public class AssessmentService(
+    ModelService modelService,
+    WorkflowInstanceService workflowInstanceService,
+    IWorkflowInstanceRepository instanceRepository
+) : IAssessmentService
 {
-    public static (decimal Unrounded, float Rounded) CalculateFinalGrade(AssessmentConfiguration config,
-        IEnumerable<AssessmentPartResult> partResults)
+    public AssessmentResult GetAssessmentResult(
+        WorkflowDefinition definition,
```
dj-petersen, line 13: This method now takes a workflow definition which is calculated when the function is called. However, the other two methods in this class take the instance itself and calculate the definition internally. I think it's good to let GetAssessmentResult also take the instance instead of the definition and get it in the method itself.

UvA.Workflow/Assessments/IAssessmentService.cs
```diff
@@ -0,0 +1,23 @@
+using UvA.Workflow.Submissions;
+using UvA.Workflow.WorkflowModel;
+
+namespace UvA.Workflow.Assessments;
+
+public interface IAssessmentService
+{
+    AssessmentResult GetAssessmentResult(
+        WorkflowDefinition definition,
```
dj-petersen, line 9: Same comment as for the AssessmentService, maybe we should pass the instance here instead of the definition, just like the other methods in this service.

UvA.Workflow/Expressions/Expression.cs
```diff
@@ -32,8 +32,6 @@ public record Expression
         ["formatDate"] = new Function<DateTime?, string, string?>((d, f)
             => d?.ToString(f, CultureInfo.InvariantCulture)),
         ["if"] = new Function<bool, object?, object?, object?>((b, t1, t2) => b ? t1 : t2),
-        ["formatDate"] = new Function<DateTime?, string, string?>((d, f)
-            => d?.ToString(f, CultureInfo.InvariantCulture)),
```
dj-petersen, line 36: Why did you remove this?

dj-petersen (changes requested):

Examples/Projects/PP/Project-PP/Steps/FinalGrade.yaml
goomens: Yes, on production this works because we imported the instances with the right data. For next year we will need the rest of the workflow so the supervisor is actually set at some point

UvA.Workflow/Expressions/Expression.cs
goomens: Because it was somehow in there twice 😊

UvA.Workflow/Assessments/AssessmentService.cs
goomens: Mm the reason for this appears to be that `InstanceService.Enrich` takes only contexts and not an actual instance and I think we want to keep it that way. But then it is needed here as well

UvA.Workflow/Assessments/AssessmentService.cs
goomens: Yes, no idea why it was there

UvA.Workflow.Api/WorkflowInstances/WorkflowInstancesController.cs
goomens: Errr oops I didn't entirely mean to commit this. I used this create an export for PPLE. It is not used currently, maybe we can keep it in but I'll add a comment to clarify

Examples/Projects/PP/Project-PP/Steps/FinalGrade.yaml
goomens: No, that doesn't really make sense but it's how I used it for now... it will need updating next year but I will change it to some less silly for now

Examples/Projects/PP/Project-PP/Steps/FinalGrade.yaml
goomens: Have moved the service call to the onsubmit of this form instead of a separate action

Examples/Projects/PP/Project-PP/Steps/FinalGrade.yaml
dj-petersen: Ah okay, do we have an external course we can test this with on the test environment?

UvA.Workflow/Assessments/AssessmentService.cs
dj-petersen: Fair point, let's leave it like this then

dj-petersen (approved):

Examples/Projects/PP/Project-PP/Forms/AssessmentSupervisor.yaml
```diff
@@ -48,6 +48,6 @@ onSubmit:
         template("Supervisor: {{ AssessmentSupervisor.LayoutLanguageComments }}\nSecond reader: {{ AssessmentSecondReader.LayoutLanguageComments }}")
 
   - setProperty:
-      property: AssessmentCombined.MotivationComments
+      property: AssessmentCombined.Motivation
       value: |
```
dj-petersen, line 52: This needs to be changed too for Forms/AssessmentSecondReader

dj-petersen (changes requested):

Examples/Projects/PP/Project-PP/Steps/FinalGrade.yaml
goomens: Yes, there are some Contexts in the test environment but they are not linked by default. You can link one manually though

Examples/Projects/PP/Project-PP/Forms/AssessmentSupervisor.yaml
goomens: Oops, done

dj-petersen (approved):

PR #205, opened 2026-07-14
Feature/dn 3887

annesnour03:
Right now all the workflow YAML lives inside this repo under `Examples/` and gets baked into the Docker image, so any config change means a full app release. This PR moves that config into its own [milestones-config](https://github.com/UvA-FNWI/milestones-config) repo and loads it at runtime instead.

What it does:
- On startup the API fetches the config from GitHub (a branch/tag/commit you point it at) and builds the workflow model from that, instead of reading bundled files. Dev and tests can still point at a local folder.
- Admins get a version picker in the UI: switch between the live config, already-loaded versions, or any branch in the config repo, and preview it. There's a "Config version" panel on the Develop page to load an arbitrary ref (branch/tag/SHA) and to reload the baseline.
- A nightly rollout restart CronJob so prod re-pulls the latest config once a day.
- A small Validator CLI so the config repo's own CI can check YAML validity without the full API.
- Tests now use their own trimmed copies under `Tests/Fixtures/` instead of `Examples/`.

### To consider
On tst/acc the backend polls every 5 minutes for changes on the branch it's pointed at (prod doesn't poll, it re-pulls via the nightly restart above). Each poll resolves the ref to a commit SHA via the GitHub API and only re-downloads and re-parses the config when that SHA changed, so an unchanged branch does no extra work.

I first tried to make the poll "free" with ETags (conditional requests), assuming a 304 wouldn't count. Turns out GitHub counts a 304 the same as a normal request, so it doesn't save anything against the rate limit. What actually helps is authenticating with a token, which I've now set up.

What counts towards the request limit (60/hr/ip unauthenticated, 5000/hr with our token):
- The ref→SHA resolve on `api.github.com`, once per poll (~12/hr per pod), whether or not anything changed.
- The tarball download from `codeload.github.com` does **not** count, it's a separate host.

I've also considered fetching locally with `git`, which wouldn't hit the rate limit at all, but that means adding a git binary to the build and executing it with parameters against a checkout on disk, which I wasn't too sure about on the safety side. The plain HTTP approach felt simpler.

> [!IMPORTANT]
> After merging, we should
> 1. Revert the line with TODO to `main` https://github.com/UvA-FNWI/milestones-config/blob/main/.github/workflows/validate.yml#L18 
> 2. Changes in https://github.com/UvA-FNWI/DataNose-v2-gitops/pull/69 are needed to get this working on main once this is merged.
> 3. Also merge https://github.com/UvA-FNWI/workflow-api/pull/205


UvA.Workflow.Api/Infrastructure/WorkflowConfigLoader.cs
```diff
@@ -0,0 +1,208 @@
...
+        _shas.TryGetValue(versionKey, out var previousSha);
+        var source = await BuildSourceAsync(@ref, previousSha);
+        if (source is null)
+            return false;
+
+        var (root, tempDir, sha) = source.Value;
+        try
+        {
+            var layout = ReadLayout(root);
+            var parser = new ModelParser(new FileSystemProvider(Path.Combine(root, "Projects")));
```
goomens: Can we get rid of the hard-coded `Projects` here as well?

UvA.Workflow.Api/Versions/VersionsController.cs
```diff
@@ -8,29 +8,93 @@ namespace UvA.Workflow.Api.Versions;
 [Authorize(AuthenticationSchemes = WorkflowAuthenticationDefaults.AnyScheme)]
 public class VersionsController(
     ModelServiceResolver modelServiceResolver,
+    WorkflowConfigLoader configLoader,
     RightsService rightsService,
     ILogger<VersionsController> logger)
     : ApiControllerBase
 {
+    private const string ProjectsPrefix = "Projects/";
```
goomens: And here 😊

UvA.Workflow.Api/Infrastructure/WorkflowConfigLoader.cs
```diff
@@ -0,0 +1,208 @@
...
+            await using (var gzip = new GZipStream(stream, CompressionMode.Decompress))
+                await TarFile.ExtractToDirectoryAsync(gzip, tempDir, overwriteFiles: true);
+
+            // codeload wraps everything in a single top-level folder.
+            var root = Directory.GetDirectories(tempDir).Single();
+            return (root, tempDir, sha);
+        }
+        catch
+        {
+            DeleteTempDir(tempDir);
```
goomens: Do we not delete the temp dir if this succeeds?
(We can't do the extraction and file handling in-memory here?)

goomens (dismissed):

UvA.Workflow.Api/Infrastructure/WorkflowConfigLoader.cs
annesnour03: It is deleted on succes, but just in the caller. There is a `finally {...}` in `LoadAsync`.

I'll look to do this in-memory, I think we can do it

UvA.Workflow.Api/Infrastructure/WorkflowConfigLoader.cs
annesnour03: Yeah I thought about this, I now just removed the notion of "Projects" in general. Its now just a folder, and we can read the workflows from any of these folders

goomens (approved):

PR #207, opened 2026-07-15
Feature/dn 3817 question percentages

codure:
https://uvanose.atlassian.net/browse/DN-3817

- Moved percentage calculations to form/questions.


dj-petersen (commented):
We have to make sure this all still works with the PR from Gerrit (https://github.com/UvA-FNWI/workflow-api/pull/204) which does the whole refactor of the assessment service/helpers/controllers structure. I think double work is being done if we merge them at the same time, so one of the PRs needs to fit with the other.

codure:
I've merged and updated with the changes from https://github.com/UvA-FNWI/workflow-api/pull/204

goomens (approved):

PR #208, opened 2026-07-16
DN-3842: Add new staffmembers to workflow

dj-petersen:


UvA.Workflow/WorkflowModel/Question.cs
```diff
@@ -201,6 +201,11 @@ public class PropertyDefinition : INamed
     /// </summary>
     public bool? AllowsExternalUsers { get; set; } = false;
 
+    /// <summary>
+    /// Determines if the propertyDefinition allows the assignment of users
+    /// </summary>
+    public bool? AllowsAssignment { get; set; } = false;
```
goomens: Why do we need this? Isn't it sufficient to set rights?

UvA.Workflow/WorkflowInstances/WorkflowInstanceService.cs
```diff
@@ -129,4 +130,90 @@ public async Task UpdateProperties(
 
         await repository.Update(instance, ct);
     }
+
+    /// <summary>
+    /// Updates a single property on a workflow instance, converts the value and saves a new version in the logs.
+    /// </summary>
+    public async Task UpdateProperty(
```
goomens, line 161: How does this relate to `AnswerService.SaveAnswer`?

UvA.Workflow/Users/RightsService.cs
```diff
@@ -287,4 +287,33 @@ public async Task<bool> CanViewCollection(
         var actions = await GetAllowedActions(instance, evaluationMode, RoleAction.View);
         return actions.Any(f => f.MatchesCollection(collection));
     }
+
+    public async Task<bool> CanEditProperty(WorkflowInstance instance, string propertyName)
+    {
+        // Properties accessed via a dot-notation path (e.g. Course.StudyAdvisor) belong to a referenced/context entity, not to this workflow instance, and cannot be edited here.
```
goomens: This is true for context but not for e.g. `AssessmentSupervisor.Motivation`?

UvA.Workflow/Users/RightsService.cs
```diff
@@ -287,4 +287,33 @@ public async Task<bool> CanViewCollection(
...
+        if (propertyName.Contains('.'))
+            return false;
+
+        var allowedEditActions = await GetAllowedActions(instance, RightsEvaluationMode.RealUser, RoleAction.Edit);
```
goomens: Why `RealUser`?

UvA.Workflow.Api/WorkflowInstances/WorkflowInstancesController.cs
```diff
@@ -173,4 +175,76 @@ public async Task<ActionResult<IEnumerable<Dictionary<string, object>>>> GetInst
                 v => v.Value
             )));
     }
+
+
+    [HttpPost("{id}/properties/{property}")]
```
goomens, line 236: Do we want to use this pattern for all array-questions, instead of the current endpoint in AnswerController? Shouldn't they be in the same controller as they are very related logic?

UvA.Workflow.Api/WorkflowInstances/Dtos/WorkflowInstanceDtoFactory.cs
```diff
@@ -50,7 +50,14 @@ await instanceService.Enrich(workflowDefinition, [context],
         // Fetch versions for all steps
         var stepVersionsMap = await GetStepVersionsMap(instance, workflowDefinition.AllSteps, ct);
 
-        var relatedUsers = GetRelatedUsers(workflowDefinition, context);
+        var canEditByProperty = await Task.WhenAll(
+            workflowDefinition.RelatedUsers.Select(async r => new
+            {
+                r.Property,
+                CanEdit = await rightsService.CanEditProperty(instance, r.Property)
```
goomens: Is this call a possible performance bottleneck since we are calling it several times? (i.e. do we need some kind of 'get all properties that I can edit variant'?)

UvA.Workflow.Api/Users/ExternalUserEmailUpdateService.cs
```diff
@@ -80,6 +100,39 @@ private IEnumerable<QuestionContext> GetMatchingUserQuestionContexts(WorkflowIns
...
 
+    /// <summary>
+    /// For user properties that are not part of a form
+    /// </summary>
+    private IEnumerable<PropertyDefinition> GetMatchingInstanceOnlyProperties(
+        WorkflowInstance instance, string userId)
+    {
+        var workflowDefinition = modelService.WorkflowDefinitions[instance.WorkflowDefinition];
+
+        // Properties that appear in at least one form are already handled by GetMatchingUserQuestionContexts
```
goomens, line 113: Do we still need that now we have this? Wouldn't a single operation suffice?

UvA.Workflow/WorkflowModel/Question.cs
dj-petersen: This means that the property can be empty. So it shows the 'Add' button when missing and can be removed.

dj-petersen: I have removed it from the yaml and made it a computed field in the DTO

UvA.Workflow/Users/RightsService.cs
dj-petersen: I changed it to Request context

UvA.Workflow/Users/RightsService.cs
dj-petersen: Good point, I changed it

UvA.Workflow.Api/WorkflowInstances/Dtos/WorkflowInstanceDtoFactory.cs
dj-petersen: Good point, I changed it

UvA.Workflow.Api/Users/ExternalUserEmailUpdateService.cs
dj-petersen: The two methods are mutually exclusive and use different rights and contexts. I guess we could but I;m not sure if it makes it clearer/more efficient since the two methods are mutually exclusive and use different rights and contexts.

UvA.Workflow.Api/WorkflowInstances/WorkflowInstancesController.cs
dj-petersen: What they do is related but everything in the answerController lives inside a submission context whereas the properties live on instance level.

UvA.Workflow/WorkflowInstances/WorkflowInstanceService.cs
dj-petersen: A big part of the logic/working is the same. The big difference is that AnswerService.SaveAnswer works inside a submission/form context. So this also handles file questions, only journals when if the form has been submitted, and checks dependent questions and returns those. UpdateProperty is more an admin path that edits the property directly and doesn't handle file questions.

UvA.Workflow.Api/WorkflowInstances/WorkflowInstancesController.cs
dj-petersen: We should probably do this array behaviour also for answers in forms, but I guess that's part of the 'handle multiple users in form input' ticket? https://uvanose.atlassian.net/browse/DN-3916 Although I could also add it here.

UvA.Workflow/WorkflowInstances/WorkflowInstanceService.cs
goomens: I kind of have the feeling it might not be strange to merge them because they essentially do the same (with or without form context), but maybe that's for another ticket...

UvA.Workflow.Api/WorkflowInstances/WorkflowInstancesController.cs
goomens: Yeah, maybe it makes sense to do it there but then we also need to think about the link between doing this with or without a form context?

goomens (dismissed):

goomens (approved):

PR #209, opened 2026-07-20
Feature/dn 3852 canvas course page filter

codure:


UvA.Workflow.Api.Authentication/CanvasLti/CanvasLaunchTargetResolver.cs
```diff
@@ -38,6 +41,52 @@ public async Task<string> ResolveTarget(User user, CanvasLaunchInfo launchInfo,
...
+            .Distinct(StringComparer.OrdinalIgnoreCase)
+            .Select(name => modelService.WorkflowDefinitions.GetValueOrDefault(name))
+            .Where(definition => definition?.Screens is { Count: > 0 })
+            .Take(2)
+            .ToArray();
+
+        if (definitions.Length != 1)
+            return options.TeacherTarget;
+
+        var definition = definitions[0]!;
```
annesnour03, line 59: This selects a potentially nondeterministic screen, because the ordering in unspecified. Should we maybe order by name so its always the same?

annesnour03 (dismissed):
1 minor comment

UvA.Workflow.Api.Authentication/CanvasLti/CanvasLaunchTargetResolver.cs
codure: I've added OrderBy to the definitions to sort by name 👍

goomens (approved):

PR #210, opened 2026-07-21
DN-3902 Added FindOpenStep for broken currentStep instances.

codure:


goomens (commented):
Doesn't this duplicate the logic in `InstanceService.UpdateCurrentStep`?

UvA.Workflow/WorkflowInstances/InstanceService.cs
```diff
@@ -123,15 +124,7 @@ public async Task UpdateCurrentStep(WorkflowInstance instance, CancellationToken
...
-        string? targetStep = null;
-        foreach (var step in workflowDefinition.FlattenedSteps)
-        {
-            if (step.Condition.IsMet(context) && !step.HasEnded(context))
-            {
-                targetStep = step.Name;
-                break;
-            }
-        }
+        var targetStep = modelService.FindOpenStep(instance, context)?.Name;
```
codure, line 127: @goomens I moved the code from UpdateCurrentStep to FindOpenStep in the modelservice so they are both using the same code.

goomens (approved):

PR #213, opened 2026-07-23
DN-3913: Make useful links dynamic 

dj-petersen:
Config PR: https://github.com/UvA-FNWI/milestones-config/pull/2

### DIFF
--- a/UvA.Workflow/Users/RightsService.cs
+++ b/UvA.Workflow/Users/RightsService.cs
@@ -83,6 +83,13 @@ public class RightsService(
 
     public async Task<string[]> GetViewerRoles(WorkflowInstance instance, CancellationToken ct = default)
     {
+        var impersonatedRoleName = await impersonationContextService.GetImpersonatedRole(instance, ct);
+        if (!string.IsNullOrWhiteSpace(impersonatedRoleName))
+        {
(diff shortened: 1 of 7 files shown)


UvA.Workflow/Users/RightsService.cs
```diff
@@ -83,6 +83,13 @@ public async Task<Domain_Action[]> GetAllowedActions(string? workflowDefinition,
 
     public async Task<string[]> GetViewerRoles(WorkflowInstance instance, CancellationToken ct = default)
     {
+        var impersonatedRoleName = await impersonationContextService.GetImpersonatedRole(instance, ct);
```
goomens, line 86: Why do we need this here?

goomens (approved):

UvA.Workflow/Users/RightsService.cs
dj-petersen: This method did not take into account whether you were using the impersonator. Therefore, it was not possible to see the right things when using the impersonator. (The viewer roles are currently also used in the frontend for the IsStudent boolean to distinguish in the UI.). Now when you impersonate you see the view that user sees, including the UI changes and the resources.

This behavior now matches the `GetAllowedActionsForRequestContext `method that is used in most/all other places.

PR #215, opened 2026-07-28
Feature/dn 3914 personal overview

codure:
https://uvanose.atlassian.net/browse/DN-3914
Uses: https://github.com/UvA-FNWI/milestones-config/pull/3
Features:
- Added PreProcess "RoleBindingMap" to the ModelParser. This creates a mapping of every user and [user] per definition with identifiers, properties and translations.
- Added 'Personal' controller and service for requesting a list of instances that the user is assigned to in for every role.
- Cleaned up ScreenDataService and the query builder by implementing the new roleBindings.


UvA.Workflow.Api/Personal/PersonalInstanceService.cs
```diff
@@ -0,0 +1,189 @@
...
+            return new PersonalInstancesDto([], []);
+
+        var directBindings = modelService.RoleBindings.All
+            .Where(binding => binding.Source == RoleBindingSource.Direct)
+            .ToArray();
+
+        if (directBindings.Length == 0)
+            return new PersonalInstancesDto([], []);
+
+        var userFilter = BuildUserFilter(directBindings, userId);
```
annesnour03: This now kind of bypasses the view-rights model. The filter does not check for `View` rights, this way we leak data to the frontend about the instances to users that may not have access to this data.

UvA.Workflow.Api/Personal/PersonalInstanceService.cs
```diff
@@ -0,0 +1,189 @@
...
+        var instances = (await workflowInstanceRepository.GetByFilter(userFilter, ct)).ToArray();
+        var courseNames = await GetCourseNames(instances, ct);
+
+        var instanceDtos = instances
+            .Select(instance => CreateDto(instance, userId, courseNames))
+            .Where(dto => dto.Roles.Length > 0)
+            .OrderByDescending(dto => dto.CreatedOn)
+            .ThenByDescending(dto => dto.Id)
+            .ToArray();
+        var roles = instanceDtos
```
annesnour03, line 34: Did we address how we were going to make sure we didn't have the case where you first have:
"You are second reviewer for"
and then
"You are reviewer for"

annesnour03 (changes requested):
Thanks for putting this together. I've posted comments cover the specific behavior, but my broader concern is the introduction of `RoleBindingMap`. I understand what it is trying to solve, but it creates another representation of role information, including precompiled indexes and inherited-role handling, while this ticket only needs direct affiliations. The workflow definitions already describe direct roles through their `User` properties, and `ObjectContext.Get` already converts those properties to `InstanceUser/InstanceUser[]`. This was previously done with `definition.Properties
            .Where(p => p.DataType == DataType.User)`

Would you be open to simplifying this by:

- building the Mongo filter from direct User properties on the relevant project
  definitions

- using ObjectContext.Get to determine the user’s roles and collect Student/Staff
  values;

I think that would meet the ticket more directly, reduce the amount of new machinery, and keep the workflow model as the source of truth. The `RoleBindingMap` may make sense later if several features genuinely need a compiled role index, but it feels bigger than necessary for this change.

UvA.Workflow.Api/Personal/PersonalInstanceService.cs
codure: I think we discussed it briefly, but don't remember a resolution to the "problem".
For now you will see the same instance for both roles.
image

UvA.Workflow.Api/Personal/PersonalInstanceService.cs
codure: YEP! completely skipped/missed that part... I'm going to revert the "RoleBindingMap" idea and implement the same logic from Screen/InstanceAuthorizationFilter for the rights per instance/role.

codure:
@annesnour03 I've removed the RoleBindingMap in favor of a smaller query builder based on the properties, with an added 'View' rights check. (I've also reverted the Screens authorization back to the original)

UvA.Workflow.Api/Personal/PersonalInstanceService.cs
annesnour03: Fixed with order I assume now

annesnour03 (approved):

PR #216, opened 2026-07-28
Feature/dn 3886 pick substeps

annesnour03:
DN-3886

## Summary

This adds support for overriding individual substeps in an inherited
workflow without having to redefine the complete parent step hierarchy.

For example, a child workflow can override Review or one of its descendants
while the remaining step configuration is inherited from the parent.

## What changed

- Child steps are merged with matching parent steps.
- Inherited parent steps are relinked to overridden child descendants.
- Explicit YAML values now win, including default values such as Sequential,
  Normal, false, [], and null.

- Multi-level inheritance is processed parent-first and inheritance cycles
  are rejected.

- Collection inheritance was made consistent for fields, properties, events,
  related users, groups, form pages, resources, and resource items.

## Inheritance rules

The general contract is:

- Omitted property: inherit from the parent.
- Explicit property: use the child value, including default values or null.
- Named collection: merge parent and child entries by name/key.
- Matching collection entry: the child entry wins.
- Empty collection: clear the complete inherited collection.
- Individual inherited collection entries cannot currently be removed.

Resources merge by resource name, and their items merge by item name.
Declaring `items: []` clears the inherited items for that resource. A child
resource still defines its own metadata such as title and type.

### DIFF
--- a/UvA.Workflow/WorkflowModel/Inheritance.cs
+++ b/UvA.Workflow/WorkflowModel/Inheritance.cs
@@ -51,8 +51,8 @@ public partial class ModelParser
         foreach (var screen in source.Screens.Where(s => !target.Screens.Contains(s.Name)))
             target.Screens.Add(screen);
 
-        if (!target.Declared("globalActions"))
-            target.GlobalActions = source.GlobalActions.Select(a => a.Clone()).ToList();
+        if (!Cleared(target, "globalActions", target.GlobalActions.Count))
+            target.GlobalActions = MergeActions(target.GlobalActions, source.GlobalActions);
 
         // Global and overridden-step actions are registered from the merged definition.
         var sourceGlobals = source.GlobalActions.ToHashSet();
@@ -147,6 +147,13 @@ public partial class ModelParser
         return result.ToArray();
     }
 
+    private static List<Action> MergeActions(List<Action> target, List<Action> source)
+    {
+        var targetNames = target.Select(a => a.Name).Where(n => n != null).ToHashSet();
+        return source.Where(a => a.Name == null || !targetNames.Contains(a.Name))
+            .Select(a => a.Clone()).Concat(target).ToList();
+    }
+
     private void ApplyInheritance(Form target, Form source)
     {
         if (Cleared(target, "pages", target.Pages.Count))
@@ -172,8 +179,8 @@ public partial class ModelParser
             foreach (var ev in source.Events.Where(e => !target.Events.Contains(e.Name)))
                 target.Events.Add(ev.Clone());
 
-        if (!target.Declared("actions"))
-            target.Actions = source.Actions.Select(a => a.Clone()).ToList();
+        if (!Cleared(target, "actions", target.Actions.Count))
+            target.Actions = MergeActions(target.Actions, source.Actions);
     }
 
     private void ApplyInheritance(SendMessage target, SendMessage source)
--- a/UvA.Workflow/WorkflowModel/Inheritance.cs
+++ b/UvA.Workflow/WorkflowModel/Inheritance.cs
@@ -90,7 +90,9 @@ public partial class ModelParser
                 .Concat(target.RelatedUsers)
                 .ToArray();
 
-        target.RelatedUserGrouping = MergeRelatedUserGrouping(target.RelatedUserGrouping, source.RelatedUserGrouping);
+        if (!Cleared(target, "relatedUserGrouping", target.RelatedUserGrouping?.Groups.Length ?? 0))
+            target.RelatedUserGrouping =
+                MergeRelatedUserGrouping(target.RelatedUserGrouping, source.RelatedUserGrouping);
         if (!Cleared(target, "resources", target.Resources.Length))
             target.Resources = MergeResources(target.Resources, source.Resources);
     }
--- a/UvA.Workflow/WorkflowModel/Inheritance.cs
+++ b/UvA.Workflow/WorkflowModel/Inheritance.cs
@@ -2,13 +2,17 @@ namespace UvA.Workflow.WorkflowModel;
 
 public partial class ModelParser
 {
+    // Explicit empty lists clear; omitted lists inherit or merge.
+    private static bool Cleared(IDeclaredKeys target, string key, int count) => target.Declared(key) && count == 0;
+
     private void ApplyInheritance(WorkflowDefinition target, WorkflowDefinition source)
     {
         target.Parent = source;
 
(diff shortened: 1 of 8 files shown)


UvA.Workflow/WorkflowModel/Inheritance.cs
```diff
@@ -2,31 +2,38 @@ namespace UvA.Workflow.WorkflowModel;
...
                 ApplyInheritance(targetForm, sourceForm);
             else
                 target.Forms.Add(sourceForm.Clone());
         }
 
-        foreach (var property in source.Properties.Where(p =>
-                     !target.Properties.Contains(p.Name) && !target.AllSteps.SelectMany(s => s.Properties)
-                         .Select(s => s.Name).Contains(p.Name)))
-            target.Properties.Add(property);
+        if (!Cleared(target, "properties", target.Properties.Count))
```
goomens: I don't think you're ever going to want to inherit and then clear all properties?

In general: could we maybe have it pass `d => d.Properties` here instead of the last two arguments for strong typing and more fool-proofing?

UvA.Workflow/WorkflowModel/Inheritance.cs
```diff
@@ -36,41 +43,58 @@ private void ApplyInheritance(WorkflowDefinition target, WorkflowDefinition sour
...
         foreach (var role in Roles)
-        foreach (var action in role.Actions.Where(a => a.WorkflowDefinition == source.Name).ToArray())
+        foreach (var action in role.Actions
+                     .Where(a => a.WorkflowDefinition == source.Name
+                                 && !sourceGlobals.Contains(a)
+                                 && !a.Steps.Any(declaredStepNames.Contains))
+                     .ToArray())
         {
             var newAction = action.Clone();
             newAction.WorkflowDefinition = target.Name;
```
goomens, line 73: Don't we need this for the global actions as well?

UvA.Workflow/WorkflowModel/Inheritance.cs
```diff
@@ -36,41 +43,58 @@ private void ApplyInheritance(WorkflowDefinition target, WorkflowDefinition sour
...
+        if (!Cleared(target, "events", target.Events.Count))
+            foreach (var ev in source.Events.Where(e => !target.Events.Contains(e.Name)))
+                target.Events.Add(ev.Clone());
 
-        foreach (var screen in source.Screens)
+        foreach (var screen in source.Screens.Where(s => !target.Screens.Contains(s.Name)))
             target.Screens.Add(screen);
 
+        if (!target.Declared("globalActions"))
+            target.GlobalActions = source.GlobalActions.Select(a => a.Clone()).ToList();
```
goomens: No merging here?

UvA.Workflow/WorkflowModel/Inheritance.cs
```diff
@@ -36,41 +43,58 @@ private void ApplyInheritance(WorkflowDefinition target, WorkflowDefinition sour
...
-            .ToArray();
-
-        target.RelatedUsers = source.RelatedUsers
-            .Where(sourceRelatedUser => target.RelatedUsers.All(targetRelatedUser =>
-                targetRelatedUser.Property != sourceRelatedUser.Property))
-            .Concat(target.RelatedUsers)
-            .ToArray();
-        target.RelatedUserGrouping = MergeRelatedUserGrouping(target.RelatedUserGrouping, source.RelatedUserGrouping);
-        target.Resources = MergeResources(target.Resources, source.Resources);
+        if (!target.Declared("steps")) target.StepNames = source.StepNames;
```
goomens: This might also benefit from an expression-based pattern

UvA.Workflow/WorkflowModel/Inheritance.cs
```diff
@@ -124,12 +149,31 @@ private static Resource[] MergeResources(Resource[] target, Resource[] source)
...
+        if (!target.Declared("resultsType")) target.ResultsType = source.ResultsType;
+        if (!target.Declared("children")) target.ChildNames = source.ChildNames;
+
+        // Reset-parent processing mutates event definitions, so inherited events cannot be shared.
+        if (!Cleared(target, "events", target.Events.Count))
+            foreach (var ev in source.Events.Where(e => !target.Events.Contains(e.Name)))
+                target.Events.Add(ev.Clone());
+
+        if (!target.Declared("actions"))
+            target.Actions = source.Actions.Select(a => a.Clone()).ToList();
```
goomens: No merging?

UvA.Workflow/WorkflowModel/Inheritance.cs
annesnour03: Good catch. I changed global actions to merge when the child supplies a non-empty list.

UvA.Workflow/WorkflowModel/Inheritance.cs
annesnour03: Yes also changed it here, same behavior as global actions

UvA.Workflow/WorkflowModel/Inheritance.cs
annesnour03: We do this for the global actions later in  `UvA.Workflow/WorkflowModel/ModelParser.cs:117`

```c#
foreach (var action in definition.GlobalActions.Except(definition.Parent?.GlobalActions ?? []))
{
    action.WorkflowDefinition = definition.Name;
    foreach (var role in action.Roles)
        Roles.Get(role).Actions.Add(action);
}
```

UvA.Workflow/WorkflowModel/Inheritance.cs
annesnour03: I don’t have a concrete use case for clearing all properties today, but I kept it to make the inheritance contract consistent across collections:
- omitted: inherit everything;
- empty list: clear everything;
- non-empty list: merge.

The remaining limitation is that we cannot remove one specific inherited item yet.

Do you think this contract matches our usecases? Or should properties behave different than other collections?

I’ve also changed the presence checks to use typed property expressions, which removes the fragile YAML-key and count pairing.

UvA.Workflow/WorkflowModel/Inheritance.cs
annesnour03: Implemented!

goomens (approved):

UvA.Workflow/WorkflowModel/Inheritance.cs
goomens: Let's start out with this, we can tweak it later

PR #217, opened 2026-07-30
DN-3840: Show photos from Canvas

dj-petersen:
uses this new version of the LTI-consumer package https://github.com/uva/LTI-consumer/pull/8


UvA.Workflow/Users/InstanceUser.cs
```diff
@@ -19,6 +19,10 @@ public class InstanceUser
     [BsonIgnoreIfNull]
     public string? PreferredLanguage { get; set; }
 
+    [BsonElement("Picture")]
+    [BsonIgnoreIfNull]
+    public string? Picture { get; set; }
```
goomens, line 24: This is an url I think? Should we call it `PictureUrl`?

UvA.Workflow/WorkflowInstances/InstanceService.cs
```diff
@@ -107,6 +107,18 @@ public async Task Enrich(WorkflowDefinition workflowDefinition,
                         workflowDefinition.AllSteps.GetOrDefault(stepName)?.DisplayTitle ?? stepName;
             }
         }
+
+        // Enrich Picture on embedded instance user snapshots from the live users table
```
goomens: Is this a logical place to do this? Do we always want to do this? Most of the time a context won't need users or pictures I guess

goomens (commented):
Question: is there a way to get picture urls from Canvas without the user logging in via Canvas? If we would want to do some bulk operation via the API?

UvA.Workflow/Users/InstanceUser.cs
dj-petersen: Yes it's the url of the user avatar. I used Picture to match LtiPrincipal of the canvas response. But I agree that maybe PictureUrl would be clearer.

UvA.Workflow/WorkflowInstances/InstanceService.cs
dj-petersen: Yeah I don't really know what a good place would be. Now it runs every time an instance is opened so you have the updated pictures. Do you have another suggestion?

dj-petersen:
> Question: is there a way to get picture urls from Canvas without the user logging in via Canvas? If we would want to do some bulk operation via the API?

This currently uses the `picture` field which is returned in the login call. There is also an `avatar_url` field in the `/users/{user_id}/profile` endpoint. I guess we could access with an admin token if we know/ask the user id used by canvas.

dj-petersen:
I'm now syncing all the info for users to the instances. Because I guess we don't want stale data in the instances. But that also means the name of the person updates to the name in Canvas. Do we want that as the source of truth? Or only for pictures?

goomens:
> I'm now syncing all the info for users to the instances. Because I guess we don't want stale data in the instances. But that also means the name of the person updates to the name in Canvas. Do we want that as the source of truth? Or only for pictures?

It should always be the same but maybe best not to rely on Canvas because not all users come from Canvas. So let's use an another single source of truth for that

dj-petersen:
> > I'm now syncing all the info for users to the instances. Because I guess we don't want stale data in the instances. But that also means the name of the person updates to the name in Canvas. Do we want that as the source of truth? Or only for pictures?
> 
> It should always be the same but maybe best not to rely on Canvas because not all users come from Canvas. So let's use an another single source of truth for that

So do we want to sync the email address or the organisation?

goomens:
No, let's not use Canvas for any of that

UvA.Workflow/Users/UserService.cs
```diff
@@ -215,26 +227,68 @@ public async Task<IEnumerable<UserSearchResult>> FindUsers(string query, bool in
         }
     }
 
-    public async Task UpdatePicture(string username, string? picture, CancellationToken ct = default)
+    /// <summary>
+    /// Finds all instances containing this user in any user-type property and replaces the
+    /// fields in the embedded snapshot with the value from the current User state.
+    /// </summary>
+    public async Task SyncUserInInstances(User user, string[] fields, CancellationToken ct)
```
goomens: For strong typing, do we maybe want to use expressions here instead of strings?

UvA.Workflow/Users/UserService.cs
```diff
@@ -215,26 +227,68 @@ public async Task<IEnumerable<UserSearchResult>> FindUsers(string query, bool in
...
+        var filter = Builders<WorkflowInstance>.Filter.Or(
+            userPropertyNames.Select(name =>
+                Builders<WorkflowInstance>.Filter.Eq($"Properties.{name}._id", userId)));
+
+        var instances = await instanceRepository.GetByFilter(filter, ct);
+
+        foreach (var instance in instances)
+        {
+            if (SyncUserInInstance(instance, user, fields))
+                await instanceRepository.Update(instance, ct);
```
goomens: I think we want to avoid using this whenever possible because it overwrites the full instance, which can cause very strange effects when operations are performed simultaneously. Can we use a property-level update here?

dj-petersen: I've made some changes

UvA.Workflow/Users/UserService.cs
dj-petersen: I've made some changes to fix this

PR #218, opened 2026-07-30
DN-3901 refactor(admin): add admin endpoint; PropertyChangeEntry and …

annesnour03:
…AnswerService for improved property journaling
DN-3901


UvA.Workflow.Api/WorkflowDefinitions/Dtos/WorkflowDefinitionDto.cs
```diff
@@ -11,7 +11,9 @@ public record WorkflowDefinitionDto(
     string? InheritsFrom,
     bool IsEmbedded,
     string[] Screens,
-    bool CanCreateInstance
+    bool CanCreateInstance,
+    // True when the definition declares steps; instance steps may be condition-filtered.
+    bool HasSteps
```
dj-petersen: Since this value is only used for checking if the workflow definition is property only maybe the name isPropertyOnly would be more descriptive

UvA.Workflow.Api/WorkflowInstances/WorkflowInstancesController.cs
```diff
@@ -85,6 +87,91 @@ public async Task<ActionResult<WorkflowInstanceDto>> GetById(string id, [FromQue
...
+
+        // The admin view ignores property conditions and visibility.
+        var properties = definition.Properties
+            .Select(p => QuestionDto.Create(p, context, totalWeight: 0))
+            .ToArray();
+
+        return Ok(new InstancePropertiesDto(properties, GetPropertyValues(instance, definition)));
+    }
+
+    private static Dictionary<string, JsonElement?> GetPropertyValues(WorkflowInstance instance,
```
dj-petersen: I think this method belongs in a service instead of a controller

UvA.Workflow.Api/WorkflowInstances/WorkflowInstancesController.cs
```diff
@@ -85,6 +87,91 @@ public async Task<ActionResult<WorkflowInstanceDto>> GetById(string id, [FromQue
...
+
+        if (!await rightsService.Can(instance, [RoleAction.ViewAdminTools], RightsEvaluationMode.RealUser))
+            return Forbidden();
+
+        var parts = path.Split('.');
+        if (parts.Length > 2)
+            return BadRequest("PropertyPathTooDeep",
+                $"Cannot edit '{path}': only properties nested at most one level deep can be saved.");
+
+        var property = modelService.GetQuestion(instance, parts);
```
dj-petersen: Is a property always a question?

UvA.Workflow.Api/WorkflowInstances/WorkflowInstancesController.cs
```diff
@@ -85,6 +87,91 @@ public async Task<ActionResult<WorkflowInstanceDto>> GetById(string id, [FromQue
...
+    /// </summary>
+    [HttpPost("{id}/properties/{path}")]
+    public async Task<ActionResult> SaveProperty(string id, string path,
+        [FromBody] SaveInstancePropertyRequest input, CancellationToken ct)
+    {
+        var instance = await repository.GetById(id, ct);
+        if (instance == null)
+            return WorkflowInstanceNotFound;
+
+        if (!await rightsService.Can(instance, [RoleAction.ViewAdminTools], RightsEvaluationMode.RealUser))
```
dj-petersen: Right now it's admin only, but there can be edit rights on property level. Shouldn't we use canEditProperty here, where we add a short circuit if the person has admin rights? That makes it more reusable in the future.

dj-petersen (changes requested):

UvA.Workflow.Api/WorkflowDefinitions/Dtos/WorkflowDefinitionDto.cs
annesnour03: Sure, good point, I'll change that.

UvA.Workflow.Api/WorkflowInstances/WorkflowInstancesController.cs
annesnour03: Agreed. I moved `GetPropertyValues` into `InstanceService`. The controller now only builds the response.

UvA.Workflow.Api/WorkflowInstances/WorkflowInstancesController.cs
annesnour03: No. The existing `GetQuestion` helper was misleading because it resolves any `PropertyDefinition`. I renamed it to `GetProperty` and updated all callers.

UvA.Workflow.Api/WorkflowInstances/WorkflowInstancesController.cs
annesnour03: Agreed. `SaveProperty` now uses `CanEditProperty`. It short-circuits for unrestricted real-user edit rights, so admins retain access while impersonating, then falls back to form/property permissions. I did this inside the actual rights check instead of in the controller.

dj-petersen (approved):
Looks good!

PR #224, opened 2026-08-05
DN-3917: Handle email edit for external users in pending state

dj-petersen:


UvA.Workflow.Api/Users/Dtos/CreateExternalUserDto.cs
```diff
@@ -3,4 +3,5 @@ namespace UvA.Workflow.Api.Users.Dtos;
 public record CreateExternalUserDto(
```
goomens: Does the `Create` need to go from the name since it is also used to update?

UvA.Workflow.Users.EduId/EduIdUserService.cs
```diff
@@ -30,27 +30,30 @@ public async Task<UserSearchResult> CreateOrUpdateExternalUser(
...
                 ExternalUserCreationFailureReason.InternalEmailAddress,
                 "Internal email address");
-        }
 
         var resolvedDisplayName = string.IsNullOrWhiteSpace(displayName)
             ? trimmedEmail
             : displayName.Trim();
-        var existingUser = await userRepository.GetByEmail(trimmedEmail, ct);
+
+        var existingUser = userId != null ? await userRepository.GetById(userId, ct) : null;
```
goomens: Is it safe to no longer match by email address here? Is the case of an existing user with the email address handled elsewhere?

goomens (dismissed):

UvA.Workflow.Api/Users/Dtos/CreateExternalUserDto.cs
dj-petersen: I've renamed it

UvA.Workflow.Users.EduId/EduIdUserService.cs
dj-petersen: It should not be possible but I've added the getbyemail as a fallback just in case

goomens (dismissed):

goomens (approved):

PR #225, opened 2026-08-06
Feature/dn 3925 resubmit status indicator

codure:
ticket: https://uvanose.atlassian.net/browse/DN-3925
milestones-config branch: ```feature/DN-3925-resubmit-status-indicator```

- Made step 'progress' an array with conditions (just like headerStatus)
- Removed double instance 'getById' in WorkflowInstanceDtoFactory
- Added property projections to Screens for progress based on properties


UvA.Workflow.Api/Screens/ScreenDataService.cs
```diff
@@ -133,15 +149,61 @@ private async Task<List<ObjectContext>> LoadData(Screen screen, string workflowD
...
         // Add related properties as needed
-        await instanceService.Enrich(modelService.WorkflowDefinitions[workflowDefinition],
-            contexts, screen.Columns.SelectMany(c => c.Properties), ct, false);
+        await instanceService.Enrich(definition,
+            contexts, screen.Columns.SelectMany(c => c.Properties).Concat(progressLookups), ct, false);
 
         return contexts;
     }
 
+    private static void AddEventActivity(
```
annesnour03: Could we handle this in `ObjectContext.Create(WorkflowDefinition, rawData)` instead?

That method already reads the Events document. It could also calculate which events are active, remove dates for suppressed events, and add the EventActive values. This would make both ways of creating an `ObjectContext` behave the same.

If we do that, we can remove:

- AddEventActivity
- IsEventContextLookup
- The hasProgressColumn check inside rawData.Select

`AddLookupToProjection` can then keep using `WorkflowDefinition.GetKey`, including for lookups inside functions. We only need to keep hasProgressColumn when deciding whether to load and enrich progress properties.

Something like this could live in the raw context builder:
```csharp
foreach (var ev in events.Values)
    {
    var eventKey = ev.Id + "Event";
    var activeKey = ev.Id + "EventActive";
    var isActive = EventSuppressionHelper.IsEventActive(ev.Id, instance, workflowDefinition);
    
    dict.Remove(eventKey);
    dict[activeKey] = isActive;
    
    if (isActive && ev.Date != null)
        dict[eventKey] = ev.Date;
    }
```
This also removes the fragile `[..^11]` and `[..^5]` checks. An event date may be loaded twice in some cases, but the context builder will normalize it, which seems simpler than maintaining separate event-lookup logic in `ScreenDataService`.

UvA.Workflow.Api/Screens/ScreenDataService.cs
```diff
@@ -133,15 +149,61 @@ private async Task<List<ObjectContext>> LoadData(Screen screen, string workflowD
...
+            Properties = [],
+            Events = events
+        };
+
+        foreach (var eventId in events.Keys)
+            context.Values[eventId + "EventActive"] =
+                EventSuppressionHelper.IsEventActive(eventId, instance, definition);
+    }
+
+    private static bool IsEventContextLookup(Lookup lookup, WorkflowDefinition definition)
```
annesnour03: See above comment, I think this can be removed...

UvA.Workflow/WorkflowModel/Step.cs
```diff
@@ -32,9 +32,21 @@ public record Icon
 
 public record ProgressInformation
```
annesnour03: `headerStatus` allows the short form:
`event: RejectSubject`

Progress requires:
```yaml
condition:
    event: RejectSubject
```

Was that difference intentional? Since these features are very similar, using the same syntax would make the config easier to understand.

UvA.Workflow.Api/Screens/ScreenDataService.cs
```diff
@@ -133,15 +149,61 @@ private async Task<List<ObjectContext>> LoadData(Screen screen, string workflowD
...
+            var context = modelService.CreateContext(workflowDefinition, row);
+            if (hasProgressColumn)
+                AddEventActivity(context, row, definition);
+            return context;
+        }).ToList();
 
         // Add related properties as needed
-        await instanceService.Enrich(modelService.WorkflowDefinitions[workflowDefinition],
-            contexts, screen.Columns.SelectMany(c => c.Properties), ct, false);
+        await instanceService.Enrich(definition,
```
annesnour03, line 151: This makes property-based progress work for screens, but `PersonalInstanceService` also calls `ProgressInformationDto.Resolve` without enriching these lookups. Direct properties and events work there, but related properties such as `Course.Name` will be missing. Should both overview paths enrich the progress lookups the same way?

UvA.Workflow/WorkflowModel/Step.cs
```diff
@@ -32,9 +32,21 @@ public record Icon
...
+    public Condition? Condition { get; init; }
+
     public StatusColor? Color { get; init; }
     public BilingualString? Text { get; init; } = null!;
     [YamlIgnore] public BilingualTemplate? ProgressTextTemplate => field ??= BilingualTemplate.Create(Text);
+
+    [YamlIgnore]
+    public IEnumerable<Lookup> Lookups =>
+    [
+        ..Condition?.Properties ?? [],
```
annesnour03: Okay this is not really related to this PR, but I saw that this relies on `Condition.Properties`, but `greaterThanOrEqual` is missing there, can we add that?

annesnour03 (changes requested):

UvA.Workflow.Api/Screens/ScreenDataService.cs
codure: That... is a really great idea. I'm on it!

UvA.Workflow/WorkflowModel/Step.cs
codure: Not intentional. I've updated them both to inherit from a 'ConditionalOption' class so they're both processed the same now, with the added benefit of future 'conditionalOption' properties.

UvA.Workflow.Api/Screens/ScreenDataService.cs
codure: I've refactored the InstanceService to allow multiple definition types as 'batches'.
Both ScreenData and PersonalInstance now share the same enrichment implementation.

UvA.Workflow/WorkflowInstances/InstanceService.cs
```diff
@@ -55,63 +60,104 @@ void Collect(IEnumerable<PropertyDefinition> properties, string[] prefix)
...
-            .Cast<PropertyLookup>()
-            .Distinct()
-            .Where(p => p.Parts.Length > 1)
-            .Where(p => workflowDefinition.Properties.GetOrDefault(p.Parts[0])?.DataType == DataType.Reference)
-            .GroupBy(p => p.Parts[0]);
-        foreach (var referenceProperty in referenceProperties)
+    /// <summary>
+    /// Populates references for multiple workflow definitions with a single repository query.
+    /// </summary>
+    public async Task Enrich(
```
annesnour03, line 93: Can we maybe document inline/functions so that this extended function (and helpers) easier to follow?

codure: I will add some! no problem

UvA.Workflow/WorkflowModel/ModelParser.cs
```diff
@@ -360,6 +360,16 @@ private void PreProcess(Step step, WorkflowDefinition workflowDefinition)
         PreProcess(step.Condition);
         PreProcess(step.Ends);
 
+        if (step.Progress.Count(progress => progress.EffectiveCondition == null) > 1)
```
codure, line 363: @annesnour03 This is where I added some fallback checks for the progress items.

UvA.Workflow/WorkflowInstances/InstanceService.cs
codure: I've added and updated some of the descriptions, with some examples and explanations about the 'batch' approach for merging multiple workflowdefinitions into one query.

annesnour03 (approved):

